PR #15, opened 2021-02-22
Audio-Only Bug if aborted with Ctr-C

HansMartin:
If a Youtube video is started in the Audio-Only mode (`-m`), and then aborted with Ctr-C, the Video starts to play because the "else" branch is taken as of the return code being non-zero. I added a check if the return code is equal to 4 (The return value of MPV if the media stream is canceled with Ctr-C) to prevent that. There is probably a much cleaner way to do this ;-)


pystardust:
Thanks for the fix @HansMartin

PR #17, opened 2021-02-22
Simple instruction for MacOS

Jac-Zac:
I have added instruction to make it work on MacOS. Since I use both Linux and MacOS I think it might be helpful to others too. But if you have any problem with it or wish me to change what I did I understand. Thanks for the great project anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,11 +87,21 @@ first hit Q to save position and quit mpv, then choose your format using
 
 * [youtube-dl github](https://github.com/ytdl-org/youtube-dl)
 
+### MacOS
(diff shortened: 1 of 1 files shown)


pystardust:
Thankyou @Jac-Zac

pystardust:
@Jac-Zac 
I think it is better to automate the process.
```
if [ "$(uname)" = "Darwin" ] ; then
            alias sed=gsed
fi
```

Jac-Zac:
Yes good idea that's better @pystardust

PR #22, opened 2021-02-23
Add random play

Euro20179:
I honestly don't know what the use case of this would be, I just thought it would be a cool feature.


pystardust:
Seems interesting, I testing it out looks good! nice work @Euro20179

PR #32, opened 2021-02-24
moved dep_ck ueberzug to place that works

Euro20179:
in the place this was before the program failed before it got to check if ueberzug exists, so I moved it to a place that only gets called with -t flag and before ueberzug gets called.


pystardust:
Hey I am really sorry, I patched another  PR and got confused with what to do. Cause I didn't understand what the conflicts were
@Euro20179 I understand the mistake. I will patch it. Thanks a lot

PR #33, opened 2021-02-24
Add flag to use ytfzf as yt-search

JojiiOfficial:
Added a new flag `-L` to use ytfzf as search tool, the youtube URL of the selected link will be printed out to stdout. I didn't go through all of the code so I may have missed something.

Edit: This doesn't work together with downloading thumbnails.

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -99,6 +99,7 @@ printf "     -f  <search query>    Show available formats before proceeding\n";
 printf "     -a  <search query>    Auto play the first result, no selector\n";
 printf "     -r  <search query>    Auto play a random result, no selector\n";
 printf "     -l  <search query>    Loop: prompt selector again after video ends\n";
+printf "     -L  <search query>    Prints the selected URL only, helpful for scripting\n";
 printf "  Use - instead of <query> for stdin\n";
 printf "\n"
 printf "  Option can be combines. Like\n"
@@ -199,6 +200,7 @@ preview_img () {
 
 # Opt variables
 show_thumbnails=0
+show_link_only=0
 main=0
 auto_select=0
 random_select=0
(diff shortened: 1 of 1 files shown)


JojiiOfficial:
Should work now

pystardust:
@JojiiOfficial instead of the last commit. Make thumbnail flag=0 in the start itself. That would be much cleaner what do you thing?

JojiiOfficial:
Yep you're right!

JojiiOfficial:
Nvm, then you don't see the thumbnails when you select the video

pystardust:
Hey, I just understood what you meant in the first solution. 
But there is one more problem. There will be "." comming from download meter we will need to account for that also. I am currently occupied, Ill get to it in a half a day.

JojiiOfficial:
No hurries. There are (should be) no more '.' coming from the thumbnail download meter.

pystardust:
Let, us do it in a way, such that both the thumbails and urls work together. Lets rework that part comletely ig. One option could to be to completely remove the dots and the downloading thumbs that is being printed.

JojiiOfficial:
That is exactly the way how it's working right now :)
Look at [this part](https://github.com/pystardust/ytfzf/pull/33/files#diff-a2ab70d1ccc85ac09c9540a95a34bb5fbe9a25b67ab7a2cc98ba5329b4391396R148-R163)

JojiiOfficial:
I (hopefully) fixed it. Generally a `-q` (quiet) option would be nice. This could be set automatically on `-L` to prevent the output.

pystardust:
@JojiiOfficial Nice work. Im merging this. Ill clean the if statemets into one liners using &&'s 
btw its posix not bash :laughing:

PR #37, opened 2021-02-25
Added better macOS support

sdushantha:
The support for macOS had been somewhat implemented in #17 but when testing it, it turned out that it did not work.

In this PR, I have created a function called `sed` which automatically decides which `sed` command to use depending on the user's OS.

Let me know if any changes needed to be made in this PR :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,9 +131,8 @@ first hit Q to save position and quit mpv, then choose your format using
 
 ### MacOS
 
-	brew install jq mpv youtube-dl fzf gsed
+	brew install jq mpv youtube-dl fzf
 
-> You have to install the GNU version of sed (gsed) because they are slightly different
 
 # Installation
 
--- a/ytfzf
+++ b/ytfzf
@@ -26,14 +26,6 @@ player_format="mpv --ytdl-format="
 ## misc
 FZF_DEFAULT_OPTS=""
 
-sed () {
-  if [[ "$OSTYPE" == "darwin"* ]]; then
-    gsed "${@}"
-  else
-    sed "${@}"
-  fi
-}
-
 # Adjusting size of menu 
 format_ext_menu () {
 	frac=$(((YTFZF_EXTMENU_LEN - 5 - 12)/11))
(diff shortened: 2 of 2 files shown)


pystardust:
@sdushantha 
It seems to have broken the script on linux.
One more issue is that you have used bashsims. Even after changing them to posix complaint I am getting seg faults.
We will need to look into this

pystardust:
I got the problem.
```sh
sed () {
...
}
```
sed is being aplied on  `()` and this is causing an error

shell: dash
OS: Linux  5.10.16-artix1-1 x86_64 GNU/Linux

sdushantha:
> One more issue is that you have used bashims

Sorry, I completely forgot that this script was written in POSIX sh, I'll fix that.

> sed is being aplied on `()` and this is causing an error

Does the function below work? I removed the space between `sed` and `()` and the function is now using `uname` instead of checking `$OSTYPE`:

```bash
sed() {                                   
    if [ "$(uname)" = "Darwin" ]; then           
        gsed "${@}"    
    else          
        sed "${@}"  
    fi
}
```

pystardust:
This is identical to aliasing sed. I wonder why that didn't work. 
Can you try this one edit to the code and tell me the outcome
```sh
## For MacOS users
if [ "$(uname)" = "Darwin" ]; then
	alias sed='\gsed'
fi
```
Add a backslash beforeg

sdushantha:
> This is identical to aliasing sed

That is what I thought as well but it seemed to not be working for me

> Can you try this one edit to the code and share the outcome

That unfortunately does not work for me. I get this output from `sed`:

```
sed: 1: ":a s_^.*var ytInitialDa ...": unused label 'a s_^.*var ytInitialData ?=__ ; s_;</script>.*__ ;'
```

pystardust:
well I don't think I need that label in sed anymore.
Change line 315 
```sh
		sed -E ':a s_^.*var ytInitialData ?=__ ; s_;</script>.*__ ;'
```
to
```sh
		sed -E ' s_^.*var ytInitialData ?=__ ; s_;</script>.*__ ;'
```
I think that should fix the issue. 
If this works then I think you can use standard mac os sed instead of gsed

sdushantha:
That seemed to fix the issue. I have updated the code accordingly 👍

pystardust:
@sdushantha 
Can you also add instruction for ueberzug? 
```
pip install ueberzug
```
Should work right?

sdushantha:
Unfortunately, `ueberzug` does not work on macOS. 

According to `ueberzug`'s repo, it states that
> ueberzug is a command line util which allows to display images **in combination with X11**

macOS does not use X11 and instead uses something called Aqua. But there is [`imgcat`](https://iterm2.com/documentation-images.html) which allows images to be shown on the terminal if the user uses [iTerm](https://iterm2.com/index.html) as their terminal emulator.

PR #38, opened 2021-02-25
Support for youtube url as parameter

JojiiOfficial:
For completeness, I added support for passing urls directly as parameter. You can still use `-f` to choose the format. This way you don't need an additional script to have a (more or less) full youtube player.

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -274,9 +274,17 @@ else
 	format_ext_menu
 fi
 
+ytURL=""
+regex='(https|http)://[-A-Za-z0-9\+&@#/%?=~_|!:,.;]*[-A-Za-z0-9\+&@#/%=~_|]' # 100% not stolen from stackoverflow
+if [[ $* =~ $regex ]]
+then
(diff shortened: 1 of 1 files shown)


pystardust:
@JojiiOfficial 
You have added a few bashism (the double square brackets) are not supported by posix shells. Interesting Idea, currently we are reworking main and an option to search again. Is  it fine if we add this feature once that is done?

pystardust:
We can do the same in posix shells with `grep -o`

pystardust:
@JojiiOfficial 
I have rewritten the feature posix complaint and removed conflicts with other PR

nycko123:
If i'm not mistaken, url is passed to the scrape_fn ().
If so, that doesn't work anymore. YT web search form doesn't print results when searching a specific url or video ID anymore. All 3rd party tools seems to be affected (youtube-dl,  duckduckgo, etc.). Seems like it's been done intentionally.  Why not pass the url directly to mpv?

PR #39, opened 2021-02-25
search again

Euro20179:
after the video ends if -sl is specified it will ask to make another search (-s alone does nothing)

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -276,8 +276,7 @@ fi
 
 # MAIN
 ## Scrape data and store video information ( and thumbnails download)
-[ $main -eq 0 ] && {
-
+main_fn () {
 	if [ "$*" = "-" ]; then
 		while read line; do
 			search_query="$search_query $line"
@@ -369,6 +368,7 @@ play_url () {
 }
 
 while true; do
+	main_fn $*
 	user_selection
 	play_url
 	save_before_exit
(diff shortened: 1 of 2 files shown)


pystardust:
@Euro20179 
This is a great idea. What if we make this in built with option `-l`?
So we get a search loop and then if you hit escape it asks you for a search. If you hit escape again it exits.
Even if we implement it as you say, we can make it so that loop is made 1, so that giving option ` -s ` would suffice

ytfzf
```diff
@@ -241,6 +243,8 @@ while getopts "LhDmdfxHarltU:" opt; do
 		r)
 			random_select=1
 			;;
+		s)	search_again=1
+			;;
```
pystardust, line 247: Add 
```
YTFZF_LOOP=1
```
So that it auto matically assumes loop when `-s` is given

pystardust:
Also we can make the process menu independent ( for dmenu and rofi users)

pystardust:
I have done some editing and merged

Euro20179:
> @Euro20179
> This is a great idea. What if we make this in built with option `-l`?
> So we get a search loop and then if you hit escape it asks you for a search. If you hit escape again it exits.
> Even if we implement it as you say, we can make it so that loop is made 1, so that giving option `-s` would suffice with out `-l`

I think -s without -l is good (i know you have implemented it I just feel the need to reply)

PR #40, opened 2021-02-26
added error message when stdin is used and -  is not provided

Euro20179:
stty errors go to /dev/null and if it fails gives an error message about stdin because stty seems to only fail when stdin is given and - is not given for a search query


pystardust:
Nice error catch

PR #42, opened 2021-02-27
changed instances of && and || to if statements for readability

Euro20179:


pystardust:
I think you removed the help info by mistake

Also the player_url function works on exit codes.
If first player doesnt play then seconds player will. Thats why I didn't use if there.
The rest can be modified to if.

pystardust:
@Euro20179 
If you want to know more about the script let me know.
I will explain the parts.

Euro20179:
oops, didn't see this till now, I did accidentally remove the help function, I'm going to try to redo this pull request

PR #45, opened 2021-02-28
changed && and || to if/else for readability

Euro20179:
(hopefully this time i didn't accidentally delete the helpinfo function lol)

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -247,17 +247,21 @@ scrape_fn () {
 	  -H 'accept-language: en-US,en;q=0.9' \
 	  --compressed
 	)"
-	[ -z "$yt_html" ] &&  {
-		printf "ERROR[#01]: Couldn't curl website. Please check your network and try again.\n";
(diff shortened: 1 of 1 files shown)


ytfzf
```diff
@@ -304,23 +308,23 @@ EOF
...
-	}
+	fi	
 }
 play_url () {
 	[ $YTFZF_CUR -eq 1 ] && printf "$selected_data" > "$current_file" ;
 
-	[ -n "$YTFZF_PREF" ] && {
+	if [ -n "$YTFZF_PREF" ]; then
 		eval "$player_format"\'$YTFZF_PREF\'  "\"$(printf "$urls" | tr '\n' '\t' | sed 's_\t_" "_g' )\""
-	} || {
```
pystardust, line 318: This line is needed as is
The intension here is, that if the format is not available then it will play in without prefered format. 
Leave the if statements in the play_url function. The rest is good

PR #53, opened 2021-03-01
-n flag

Euro20179:
does the selection process n times

example:
`ytfzf -aL -n2 hello`
prints
```
https://www.youtube.com/watch?v=YQHsXMglC9A
https://www.youtube.com/watch?v=mHONNcZbwDY
```

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -287,7 +287,7 @@ user_selection () {
 		selected_data="$(echo "$videos_data" | video_menu_img )"
 		stop_ueberzug 
 	else
-		selected_data="$selected_data\n$(echo "$videos_data" | video_menu )"
+		selected_data="$(echo "$videos_data" | video_menu )"
 	fi
 	
 	shorturls="$(echo "$selected_data" | sed -E -e 's_.*\|([^|]+) *$_\1_')"
--- a/ytfzf
+++ b/ytfzf
@@ -277,24 +277,18 @@ user_selection () {
 	[ $is_url -eq 1 ] && return
 	format_menu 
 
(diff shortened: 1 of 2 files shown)


JojiiOfficial:
Doesn't -L already do this?

Euro20179:
> Doesn't -L already do this?

No because if you do -L and don't want the menu screen you can't select more than 1 link

Euro20179:
don't merge this yet because there some oversights.

JojiiOfficial:
> No because if you do -L and don't want the menu screen you can't select more than 1 link
Good point. Didn't even notice the '-L' in your example query.

pystardust:
Nice idea, but doest it suffice to change
```
if [ $auto_select -eq 1 ] ; then
		    selected_data="$selected_data\n$(echo "$videos_data" | sed ${i}q | tail -n1)" ;
```
This would select the first `i` videos

ytfzf
```diff
@@ -275,24 +276,31 @@ scrape_fn () {
...
-		selected_data="$(echo "$videos_data" | shuf -n 1)"
+		selected_data="$(echo "$videos_data" | shuf -n $link_count )"
 	elif [ $show_thumbnails -eq 1 ] ; then
 		dep_ck "ueberzug"
 		start_ueberzug
 		selected_data="$(echo "$videos_data" | video_menu_img )"
 		stop_ueberzug 
 	else
-		selected_data="$(echo "$videos_data" | video_menu )"
+		selected_data="$selected_data\n$(echo "$videos_data" | video_menu )"
```
pystardust, line 290: we wouldn't need to append cause no data is being selected before this

PR #55, opened 2021-03-02
implements FZF_DEFAULT_OPTS

### LINKED ISSUE #54: FZF_DEFAULT_OPTS not being referenced
I like using the following in my default opts: `FZF_DEFAULT_OPTS="--bind=ctrl-a:select-all,ctrl-d:deselect-all,ctrl-t:toggle-all"`, but I noticed they weren't taking when I added them to the script. `$FZF_DEFAULT_OPTS` needs to be referenced in the following functions:

```diff
format_menu () {
	if [ $is_ext_menu -eq 0 ]; then
		dep_ck "fzf"
++		prompt_menu='fzf -m --bind change:top --tabstop=1 --layout=reverse --delimiter="$(printf "\t")" --nth=1,2 $FZF_DEFAULT_OPTS'
--		prompt_menu='fzf -m --bind change:top --tabstop=1 --layout=reverse --delimiter="$(printf "\t")" --nth=1,2'
```

```diff
video_menu_img () {
	title_len=400
	format_awk | fzf -m --tabstop=1 --bind change:top --delimiter="$(printf "\t")" --nth=1,2 \
		--layout=reverse --preview "sh $0 -U {}" \
++        	--preview-window "left:50%:noborder:wrap" $FZF_DEFAULT_OPTS
--        	--preview-window "left:50%:noborder:wrap" $FZF_DEFAULT_OPTS
}
```

Euro20179:
fix #54

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -23,9 +23,6 @@ thumb_dir="$YTFZF_CACHE"/thumb
 player="mpv"
 player_format="mpv --ytdl-format="
 
-## misc
-FZF_DEFAULT_OPTS=""
-
 helpinfo () {
 printf "Usage: %bytfzf %b<search query>%b\n" "\033[1;32m" "\033[1;33m" "\033[0m";
 printf "     -h                    Show this help text\n";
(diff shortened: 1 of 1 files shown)


pystardust:
Defaults args are taken by default.
The problem is some people have previews on in a different way in their default args. So i am clearning the default args in the start of the script

pystardust:
We can add some sort of a choice for people who want their fzf defaults. What do you thing?

Euro20179:
I think that there should be a way to have fzf defaults although idk how it could be implemented, this seems like a pretty simple way to do it.

Euro20179:
I did some basic testing and unless you break fzf itself with the FZF_DEFAULT_OPTS it seems to work anyway (tested `--delimiter` `--tabstop` `--bind change --nth`), I could have tested them badly because i don't really know what I'm doing lol, but it seems to work

Euro20179:
there's another option where there could be YTFZF_ENABLE_FZF_DEFAULT_OPTS or something (that variable name is kinda long) where when it's enabled fzf_default_opts works, it should be off by default.

pystardust:
Initially I didn't clear the defaults, someone on reddit pointed out that thier defaults are really absurd for ytfzf. Thats why I thought of clearning them. But yes we can do that.

PR #51, opened 2021-03-01
Changable player

westofer:
fix #50

PR #58, opened 2021-03-02
Exit earlier if dependency missing

cybarspace:
If there's a dependency missing, I don't think there's any reason to create functions and do a bunch of operations before reporting it with a crash. The CPU time and the memory footprint are minor, but it's just good practice if the crash happens before all the functions and other things are defined. Also, I think putting a `[OPTIONS]` indicator in the help screen would be nice. Many programs have it (for example, kitty, python, gcc) and it makes things clearer (where to put the flags, where to put args). Explicit is better than implicit, after all.

It's also conventional to use hyphens or underscores instead of spaces in indicators like `search-query`.

These are just suggestions. Please feel free to edit as you see fit, or close the PR if inappropriate.

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -30,29 +30,39 @@ thumb_dir="$YTFZF_CACHE"/thumb
 [ -d $thumb_dir ] || mkdir -p $thumb_dir
 
 
+# DEP CHECK
+dep_ck () {
+	for dep in "$@"; do
+		command -v "$dep" 1>/dev/null || { printf "$dep not found. Please install it.\n" ; exit 2; }
+	done
+}
+dep_ck "jq" "youtube-dl" "$YTFZF_PLAYER";
+
+
 ############################
 #       Help Texts         #
 ############################
 helpinfo () {
-printf "Usage: %bytfzf %b<search query>%b\n" "\033[1;32m" "\033[1;33m" "\033[0m";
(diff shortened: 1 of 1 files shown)


cybarspace:
Let me know if I need to edit anything, or if you have any concerns with the changes.

ytfzf
```diff
@@ -30,29 +30,39 @@ thumb_dir="$YTFZF_CACHE"/thumb
 [ -d $thumb_dir ] || mkdir -p $thumb_dir
 
 
+# DEP CHECK
+dep_ck () {
+	for dep in "$@"; do
+		command -v "$dep" 1>/dev/null || { printf "$dep not found. Please install it.\n" ; exit 2; }
+	done
+}
+dep_ck "jq" "youtube-dl" "$YTFZF_PLAYER";
```
westofer, line 39: #51 `YTFZF_PLAYER` might be not a single command, example `devour mpv` or `mpv --some-option`

cybarspace: Yeah, sorry. Didn't think about that. What I meant for was that by setting `mpv` in the dependency, one cannot use something like VLC on a system without mpv, without changing it in multiple places (sorry if that doesn't make sense) to avoid dependency errors.

Reverting the change to `dep_ck`.

PR #60, opened 2021-03-02
Revert change to dep_ck function

cybarspace:
Reverting a change I made in [PR#58](https://github.com/pystardust/ytfzf/pull/58) which incorrectly causes crashes when `YTFZF_PLAYER` has multiple words.

It'd still be nice to have a better `dep_ck` function so that mpv wouldn't be a hard dependency. It isn't, really, and one could probably use VLC or some other player to the same effect.

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -36,7 +36,7 @@ dep_ck () {
 		command -v "$dep" 1>/dev/null || { printf "$dep not found. Please install it.\n" ; exit 2; }
 	done
 }
-dep_ck "jq" "youtube-dl" "$YTFZF_PLAYER";
+dep_ck "jq" "youtube-dl" "mpv";
 
 
 ############################


Euro20179:
You could try doing this
```sh
dep_ck "${YTFZF_PLAYER%% *}"
``` 
so that way it only checks if you have the command and gets rid of any options or values, etc..

cybarspace:
> so that way it only checks if you have the command and gets rid of any options or values, etc..

We actually need the secondary values imo. Someone was using devour, so the command was `devour mpv`. In this case we'd want to check if each of these two programs is installed. We need to also drop flags, since they're not dependencies.

Maybe some way to split up the `YTFZF_PLAYER` constant by spaces, then consider each word a dependency IF they don't begin with a hyphen.

Euro20179:
That is true, maybe something like this could work
```sh
#not putting quotes around the variable splits it by word
for i in $YTFZF_PLAYER
do
         #${i#**} gets the first character of string
         if [ ${i#**} != "-" ]; then
                 dep_ck $i
         else
                 break
         fi
done
```

cybarspace:
> I don't know an actual example but you get the point

`xfce4-terminal -e mpv`

Try that loop without the else statement, so that it just skips any action for words beginning with hyphen, instead of completely breaking the loop.

Euro20179:
> so that it just skips any action for words beginning with hyphen

The problem with this is that if you have an option that doesn't require a command and instead an option it will run dep_ck on it  and it will always throw an error
again I can't think of an example but
`command --option value --option2 mpv`

cybarspace:
> The problem with this is that if you have an option that doesn't require a command and instead an option it will run dep_ck on it and it will always throw an error

Yeah, didn't think of that...

> it essentially comes down to whether or not something is a command and that's impossible to know in just code

True. Maybe hardcoding is the best way for now. Keeping it `mpv` would probably reduce dev headache.

Euro20179:
> True. Maybe hardcoding is the best way for now

Or since it's the user that's setting it they should know what they have installed right? so maybe don't check if it isn't set to mpv

pystardust:
the last line might create a problem
```
eval "$YTFZF_PLAYER" > /dev/null
```
What if the player launches with out args, suppose you run `vlc` with out any arguments, it will open the player in a new window

Yes we can dep_ck only the payer hasn't been changed, as you mentioned.
And in the end we can print an error message saying that either '
`$YTFZF_PLAYER couldn't open the url, check if $YTFZF_PLAYER is installed`

PR #61, opened 2021-03-02
config file

Euro20179:
Instead of having to use environment variables there is now a config file that overrides any environment variables

When the config file is first created it uses the environment variables that are already set, so this shouldn't affect anyone's settings

With a config file it can be much more expandable than environment variables

Yes I know it's just a sh file but it's a LOT easier to edit than the script itself


pystardust:
yes the config file option is really nice.
But do we need to automatically create a config, if not present?
Some one may not want to have a config dir in their .configs.
What if we implement it if the file exists and not create it if it doesnt

pystardust:
Cause if someone is using envariables, runs the script and then changes the env variables then the script will use the previous version that was written into the file

Euro20179:
> Cause if someone is using envariables, runs the script and then changes the env variables then the script will use the previous version that was written into the file

True, I guess we don't need to set the config file by default

PR #68, opened 2021-03-03
search filter

Euro20179:
a way to use the youtube search filter


pystardust:
Nice this is really cool. Let us test it even more before merging.

PR #73, opened 2021-03-04
Config options

Euro20179:
allows options to also be set in config
useragent is changeable in config

personally I found a very nice use for allowing options to be set in config:
when using rofi to launch ytfzf i'd need to do this
`ytfzf -D blah blah blah`
but in config i can just check if i'm launching it with rofi and set -D automatically


pystardust:
I guess you added test.json by mistake. Remove it ill merge this. Nice idea

pystardust:
The search again and loop being independent is a really nice implemetation.

PR #74, opened 2021-03-04
Code updating

Euro20179:
changes to make the code more readable
* changed some exit codes
    * exit 1 is a random error
    * exit 2 is missing parameter, missing dependency, or illegal option
* renamed format_awk to format_video_data
* renamed errinfo to print_error (also accepts arguments to print)


ytfzf
```diff
@@ -318,21 +318,19 @@ get_search_query () {
 	else
 		if [ -z "$search_query" ]; then
 			if [ $is_ext_menu -eq 1 ]; then 
-				search_query="$(printf "" | $YTFZF_EXTMENU)"
+				search_query="$(eval $YTFZF_EXTMENU)"
```
pystardust, line 321: This is needed cause if you do just 
```
ytfzf -D
```
Then you get a search prompt. Dmenu and rofi needs some stdin.

pystardust: if you do just `dmenu ` you wont get a prompt. You need some stdin.

PR #76, opened 2021-03-04
Fix #75

### LINKED ISSUE #75: Multiple sessions break thumbnails
Thumbnails do not show properly if more than one session with thumbnails enabled exists on the system.
The issue probably originates from the hard-coded variables for ueberzug (line 230-232).

Euro20179:
I have no idea what's going on with the pull requests but this fixed #75


pystardust:
we also need to manage how the images are being deleted.

Euro20179:
oh apperently this doesn't work as i thought anyway

PR #82, opened 2021-03-04
Thumbnail

Euro20179:
this fixes the issue mentioned in #75 however it comes with it's own issue, the thumbnail directory will not be removed if looping is off and fzf is escaped without selecting anything, this is because when you close fzf the whole program closes, i tried doing some things with trap but it was ineffective, I might have been looking for the wrong signals but i did 2 3 9 and 15.

(and still i have no idea what's up with the commits it was a struggle to push this to origin)


pystardust:
Again It is hard to see the changes being made to the file. Try recloning the repo and changing only the script file.

PR #85, opened 2021-03-05
thumbnail

### LINKED ISSUE #75: Multiple sessions break thumbnails

Euro20179:
this should fix #75 but there is 1 issue that i noticed, when escaping out of fzf without selecting anything it will not remove the thumbnail directory because closing out of fzf closes out of ytfzf as well

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -24,10 +24,18 @@ YTFZF_PLAYER_FORMAT="${YTFZF_PLAYER_FORMAT-mpv --ytdl-format=}"
 YTFZF_ENABLE_FZF_DEFAULT_OPTS=${YTFZF_ENABLE_FZF_DEFAULT_OPTS-0}
 [ $YTFZF_ENABLE_FZF_DEFAULT_OPTS -eq 0 ] && FZF_DEFAULT_OPTS=""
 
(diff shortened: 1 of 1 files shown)


pystardust:
Also if you have 2 sessions open and you close one. The thumbs get deleted.

Euro20179:
from my testing it works just fine after closing one of the sessions

pystardust:
when the `save_before_exit` function is runs, the  thumbs are being removed, you might have exit the session before that.

Euro20179:
it's possible, however each ytfzf instance gets it's own thumb_dir, so it shouldn't matter.

pystardust:
@Euro20179 
I have merged this to a new branch multi-session and modified it. Now the thumbnails are being deleted. Even if a session is closed with sigkill, the next time the script runs it will delete the thumbs while leaving. 
- I have changed PPID to pid `$$`
- Stored pids in a pid file in cache dir
- using only on thumb dir
- added function delete_thumbnails, which will delete thumbnails if no other session is running.

Euro20179:
nice

PR #87, opened 2021-03-05
-v, --version

Euro20179:
-v for ytfzf's version, --version for ytfzf + youtube-dl's version

change YTFZF_VERSION to whichever version is necessary

merge when you think is a good time

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -51,6 +51,8 @@ config_file="$config_dir/conf.sh"
 #source config file if exists (overrides env variables)
 [ -e $config_file ] && . "$config_file"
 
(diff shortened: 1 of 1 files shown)


pystardust:
juse parse opt `-v` instead of writing it twice

pystardust:
I added a release name `v1.0.0`

Euro20179:
> juse parse opt -v instead of writing it twice

--version prints youtube-dl's version as well, if --version did just -v, it would only print yfzf's version, they could have the same behavior but it's nice to have both options, if there are more hard dependencies, their versions could also be printed with --version

PR #89, opened 2021-03-05
Keep thumbs

Euro20179:
an option to keep the thumbnail directory

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -87,6 +87,8 @@ show_format=${show_format-0}
 link_count=${link_count-1}
 #after video ends, make another search (same as -s)
 search_again=${search_again-0}
+#after the video ends $thumb_dir is deleted, when set to 1 this prevents that from happening
+keep_thumbs=${keep_thumbs-0}
 
 #dependency check
 dep_ck () {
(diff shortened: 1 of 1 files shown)


pystardust:
How about adding it directly in the delete thumbs function to keep checking minimum.
Like adding this in the start of the delete_thumbnails function
```
[  keep_thumbs -eq 0 ] && return
```

pystardust:
And another thing to keep in mind. The thumbnails would be downloaded irrespective of the image being present already. We need to add a check for that

ytfzf
```diff
@@ -326,6 +330,11 @@ download_thumbnails () {
 			url="$(echo $line | sed -E 's/^"//;s/\.png.*/\.png/')" 
 		else
 			name="$(echo "$line" | tr -d '"')"
+			if [ -e "$thumb_dir/$name.png" ]; then
+				i=$((i + 1))
+				continue
+			fi
+
 			{ curl -s "$url" > "$thumb_dir/$name.png" && [ $show_link_only -eq 0 ] && printf "."; } &
```
pystardust, line 338: How about just onle line
```sh
[ -e "$thumb_dir/$name.png" ] || { curl -s "$url" > "$thumb_dir/$name.png" && [ $show_link_only -eq 0 ] && printf "."; } &
```
Also with the current approach, i would go to i+1 and that would change odd to even and file names would mess up.

PR #91, opened 2021-03-05
a default conf.sh

Euro20179:
documentation for each option that can be set in conf.sh

### DIFF
--- a/conf.sh
+++ b/conf.sh
@@ -12,7 +12,7 @@
 
 #the variables below can either be set here, or using export
 #variables set here will override any set by export
-#if you want to only use export you can remove the setting of the variables from here
+#if you want to only use export you can remove the variables from here
 
 #enable/disable history (enabled by default)
 YTFZF_HIST=1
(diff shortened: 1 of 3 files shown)


pystardust:
Let us try making the config file variables the same as the long opt variable.
And also lets try soucing the config before the env variables(by using different names from env var). Cause env variables are better for one time modification.
I don't thing we need another file if we document well. Maybe we can add a directory in the git repo called `docs/` or something like that and put the example config there.

Euro20179:
> And also lets try soucing the config before the env variables(by using different names from env var). Cause env variables are better for one time modification. As in it would be nice if env variables overrode the config.

I thought the same thing but i was gonna do it in a different pr but I can do it here.

> I don't thing we need another file if we document well. Maybe we can add a directory in the git repo called docs/ or something like that and put the example config there.

I think we should have some kind of documentation because I had to read through the code a lot writing that file, if we had `docs/` that could work

PR #93, opened 2021-03-06
Update flag

Euro20179:
run `sudo ytfzf --update` to update ytfzf, instead of
```
git clone https://github.com/pystardust/ytfzf
cd ytfzf
sudo make install
cd ..
rm -r ytfzf
```

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -564,6 +564,23 @@ clear_history () {
 	fi
 }
 
+update_ytfzf () {
+	#check for root privileges
+	if [ $(id -u) -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


JojiiOfficial:
I think you should not let this entire process run as root. Escalate privileges, when necessary. If you need to make it headless, you can 'unlock' sudo first.

pystardust:
We can instead just curl ytfzf from master branch and write it on script location, and change perms.
Depending on where the current ytfzf script is, we can look at the perms and if needed ask for sudo privilages.

PR #121, opened 2021-03-10
A r a subscriptions

Euro20179:
some settings in config wouldn't work eg: `history_file`
`ytfzf -Sa` will play the first video from each subscription instead of *the* first video 

note: `ytfzf -Sr` works the same as `ytfzf -Sa` the difference is that, the videos will be in a random order

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -467,12 +467,11 @@ scrape_channel () {
 	    }
 	]')"
 	
-	videos_data="$(printf "\n%s\n%s\n" "--------------------${channel_name}------------------------------------------------------" \
-		"$(get_video_data "$videos_json")" )"
+	videos_data="$(get_video_data "$videos_json" | head -${link_count} )"
 	
 	#if there aren't videos
 	[ -z "$videos_data" ] &&  { printf "No results found. Make sure the link is correct.\n"; exit 1;}
-	printf "%s" "$videos_data" >> "$tmp_video_data_file"
+	printf "$videos_data\n" >> "$tmp_video_data_file"
 	
 	[ $show_thumbnails -eq 1 ] && download_thumbnails "$videos_json"
 }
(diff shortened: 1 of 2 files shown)


ytfzf
```diff
@@ -345,18 +347,13 @@ preview_img () {
...
-			{ 
-				curl -s "$url" > "$thumb_dir/$name.png" 
-			} &
-		fi
-		i=$((i + 1))
+		url="$line"
+		name="$(echo "$url" | grep -Po 'vi/\K.*(?=/)')"
+		{ 
+			curl -s "$url" > "$thumb_dir/$name.png" 
+		} &
```
pystardust, line 356: This wouldn't impact the time. Since jq is scraping all together in the start.  
The major time improvements would come from the position of `wait` and `curl`s

In this method, we are relying on the videoID to be in the thumbnail. ( wouldn't work well with other sites).

ytfzf
```diff
@@ -468,10 +465,16 @@ scrape_channel () {
...
+	[ $select_all -eq 0 ] && videos_data="$(echo "$videos_data" | head -${link_count})"
+	if [ $auto_select -eq 1 -o $random_select -eq 1 ]; then
+	    user_selection
+	    echo "$urls" >> "$tmp_video_urls_file"
+	fi
 	
 	#if there aren't videos
 	[ -z "$videos_data" ] &&  { printf "No results found. Make sure the link is correct.\n"; exit 1;}
-	printf "%s" "$videos_data" >> "$tmp_video_data_file"
+	printf "$videos_data\n" >> "$tmp_video_data_file"
```
pystardust, line 477: Variable in printf is a bad practice

ytfzf
```diff
@@ -468,10 +465,16 @@ scrape_channel () {
 	]')"
 	
 	videos_data="$(get_video_data "$videos_json")"
+	#if select all or random select keep all the links so that they can all be selected
+	[ $select_all -eq 0 ] && videos_data="$(echo "$videos_data" | head -${link_count})"
+	if [ $auto_select -eq 1 -o $random_select -eq 1 ]; then
+	    user_selection
+	    echo "$urls" >> "$tmp_video_urls_file"
+	fi
```
pystardust, line 473: This seems to work,
I would be much cleaner if we did our usual auto-select, random select, directly on the video data. And push it through User selection with `-A`

Euro20179: idk how to make it cleaner, and keep the functionality of having it play from each subscription

pystardust:
Feature implemented

PR #123, opened 2021-03-10
error for bad -n argument

Euro20179:


pystardust:
We could make this as a function. And use it for subcount also

Euro20179:
this is on the dev branch, so i will merge it with the sub branch when this is merged.

PR #128, opened 2021-03-10
format_awk customization

Euro20179:
allows easier customization of the menu


pystardust:
Maybe we can replace awk completely with printf, like we did to thumbnail previews. We could add a custom function then like we did for thumbnails

pystardust:
Addressed in #152

PR #131, opened 2021-03-11
--install, --uninstall

Euro20179:
removes the need for Makefile
--install is it's own thing becuase --update is slower as it needs curl


pystardust:
Since it is just a cp command to install we can just add an install script in the read me.
People are accustomed to the make file, so it is better we let it be there. 
```sh
sudo curl -L "https://raw.githubusercontent.com/pystardust/ytfzf/master/ytfzf" -o /usr/bin/ytfzf
sudo chmod +x /usr/bin/tuxi
```

Euro20179:
ok

PR #138, opened 2021-03-13
Printf vs echo

Euro20179:


ytfzf
```diff
@@ -108,7 +108,7 @@ scrape="${scrape-yt_search}"
 #this could be any unique value, $$ is used because it is the most easily accessible unique value
 if [ -z "$PROC_ID" ] ; then
 	export PROC_ID="$$"
-	echo "$$" >> "$pid_file"
+	printf "$$" >> "$pid_file"
```
pystardust, line 111: This would add the pid's one after the other, rather than a new line.

pystardust (changes requested):

PR #140, opened 2021-03-14
Development merge 1.1.0

pystardust:
# Features
* Subscriptions
    * subscriptions can be set with $subscriptions_file (~/.config/ytfzf/subscriptions by default)
    * a subscription is a link to a channel's video page, eg: https://www.youtube.com/user/PewDiePie/videos (see README)
* multiple urls for input
* channel name, views, etc... has color when using -t

# Changes
* not using - for stdin gives error code of 2 instead of 1
* --version now bolds each module, and prints the version of fzf if installed

# Bug fixes
* when reading from an external menu, ytfzf threw an error when it shouldn't have
* if -t was given, it didn't check if fzf was installed, it will now
* when -tD was given, it would switch to use fzf, instead it warns about external menus not having thumbnail support then continues
* history_file and current_file didn't work in conf.sh
* issue with tac on macos
* issue with YTFZF_PLAYER not being parsed correctly

> update log by @Euro20179

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,61 +24,157 @@
 
 _These links will take you where you want to go with the snap of a finger_
 
-- [`Usage instruction`](#Usage-Instructions)
 - [`Features`](#Features)
+- [`Usage instruction`](#Usage-Instructions)
 - [`Examples`](#Examples)
 - [`Dependencies`](#Dependencies)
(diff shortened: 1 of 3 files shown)


Euro20179:
should change $YTFZF_VERISON to 1.1.0

PR #148, opened 2021-03-16
Subscription links

Euro20179:


pystardust:
So this means if the user has put the channel link instead of video page link it would still work right

Euro20179:
yes, it should, i tested with my subscriptions file

pystardust:
I have sent a PR on better checks on this branch
Also the main page for a channel has `/featured` in the end. I added a check to replace it.

PR #149, opened 2021-03-16
error, new line

Euro20179:


pystardust:
I have corrected the new line when implemeting thumbnail quality. Also the error over there will mess up url playing. ie playing a url directly using
```sh
ytfzf $url
```

Euro20179:
ok

PR #155, opened 2021-03-19
channel results

Euro20179:
too many messed up commits with the last one


pystardust:
I have created a branch for this. We could merge the development update and keep this for the next update.
 Send the PR to that one

- We would need to work on the formatting to indicate it is a channel instead of a video

Euro20179:
ok

PR #161, opened 2021-03-20
Better looking help

Jac-Zac:
I made some really small changes to make it look a bit better


pystardust:
We initially had colors in the help. But the problem with this is, when people pipe the colors into `less` then it shows garbage there. Since the help is so huge, many would use `less` to view it.

Jac-Zac:
Ah, really. wired it works for me but if it doesn't for other ok

PR #166, opened 2021-03-24
zsh complains of syntax error

Euro20179:


pystardust:
fixed in #167 along with notifications

PR #167, opened 2021-03-24
fix+feat: add --notification= option, fix zsh complain syntax

TChanhTinh:
- Add enable_noti=<1:default, 2> config that enable notification feature
- Add flag -N --notification <0,1> enable/disable notification feature
- Add notification flag -h entry
- If enable notification & thumbnails it will use thumbnail as notification icon else it will use default thumbnail icon (user defined) file at ./config/ytfzf/default_thumb.png
- Fix zsh complain syntax at download_thumbnails

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -29,6 +29,8 @@ printf "" > "$tmp_video_data_file"
 [ -z "$YTFZF_LOOP" ] && YTFZF_LOOP=${enable_loop-0}
 #enable/disable outputting current track to $current_file
 [ -z "$YTFZF_CUR" ] && YTFZF_CUR=${enable_cur-1}
+#enable/disable notification
+[ -z "$YTFZF_NOTI" ] && YTFZF_NOTI=${enable_noti-1}
 #the cache directory
 [ -z "$YTFZF_CACHE" ] && YTFZF_CACHE="${cache_dir-$HOME/.cache/ytfzf}"
 #video type preference (mp4/1080p, mp4/720p, etc..)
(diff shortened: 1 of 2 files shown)


ytfzf
```diff
@@ -151,6 +153,7 @@ basic_helpinfo () {
     printf "     -l, --loop           <search-query>    Loop: prompt selector again after video ends\n";
     printf "     -s, --search-again   <search-query>    After the video ends make another search \n";
     printf "     -L, --link-only      <search-query>    Prints the selected URL only, helpful for scripting\n";
+    printf "     -N, --notification   <0,1>  			Send notification when play video\n";
```
pystardust, line 156: It doesn't need a 0 or a 1 right?
We could just leave that part empty

pystardust: if you are making it in such a way that it takes it arguemnt, then use --notification=
and also for parse opt you would need to send the argument as `"$2"`, refer the other flags that are already been done.

ytfzf
```diff
@@ -995,13 +1018,15 @@ parse_opt () {
...
+		N)	YTFZF_NOTI=${optarg:-1} ;;
+
 		*)
 			usageinfo
 			exit 2 ;;
 	esac
 }
 
-while getopts "LhDmdfxHaArltSsvn:U:-:" OPT; do
+while getopts "LhDmdfxHaArltSsvnN:U:-:" OPT; do
```
pystardust, line 1029: here option `n` also takes a `OPTARG`, it should be `n:`

TChanhTinh: Thanks for the tips, was a little confuse at that, now flag -N now is work without arg i also change the -h description for it

ytfzf
TChanhTinh: fixed on newer commit

TChanhTinh:
Sorry for the mess, this is the first time is make a pull request, all problem you mentioned are solved.

pystardust:
np, there are few more things we could do.
Here when we select multiple videos we get a error, we would need to adjust it so that even with multiselect we get a  formatted notification.

TChanhTinh:
> np, there are few more things we could do.
> Here when we select multiple videos we get a error, we would need to adjust it so that even with multiselect we get a formatted notification.
Done in new commit, if user select more than one video, it will send notify "Add x video to play queue" & use default_thumb instead individual thumbnail

PR #169, opened 2021-03-25
Development merge 1.1.1

pystardust:


pystardust:
v1.1.1
====features====
* improved formatting: formatting can be set in conf.sh
* video notification: with flag `-N`
* thumbnail quality: `--thumbnail-quality` can be set to `0` for lower quality thumbnails which load faster
* improved subscriptions urls: channel urls could also be used in subscriptions instead of video page url

====changes====
* better memory management by using local variables
* more modular code
* use of printf instead of echo

====bug fixes====
* bad directory path in conf.sh
* zsh compliant syntax

PR #171, opened 2021-03-26
New features: store and show view dates, allow clearing partial history

ido50:
This pull request introduces two features, in two separate commits. There are detailed explanations in each commit message, but in short, the features are:

1. Store view dates in history, show when `-H` is used.
2. Allow removing videos from history that were viewed before the provided date or interval (e.g. '2 days ago') via new option `--clear-before`.

It is my hope that you will at least find the first feature useful for common users. The second feature requires the first, but I'm not sure you'll find it useful enough to include. Personally, I'd like to automate clearing of history (e.g. via a cron job), but still keep a certain amount of recent history. If you feel only the first feature should enter, I'll gladly open a PR purely for it.

These are backwards compatible changes, and I have taken care to make sure all changes are POSIX-compatible. That said, it's entirely possible that I missed something in my tests. Please read the note in the first commit message about terminal widths and let me know if any changes are necessary.

Thanks!

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -39,11 +39,18 @@ If you wish to find inspiration and see an example you can take a look at the [`
 		ytfzf -H
 		```
 
-	- Clear history
+	- Clear entire history
 
 		```sh
 		ytfzf -x
 		```
+
+    - Clear old history
+
+        ```sh
+        ytfzf --clear-before='2 days ago'
+        ytfzf --clear-before='2021-03-26 12:00'
+        ```
 
 	- History file: `~/.cache/ytfzf/ytfzf_hst`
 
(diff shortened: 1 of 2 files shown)


pystardust:
Nice implementation, but please pull requests on the development branch. We don't push directly to master.

pystardust:
As you mentioned regarding cronjobs. One could trim the top of the history file and make sure the history file has max n lines. 
 Also let us try thinking different ways to store view date. Since currently short url is being pushed to the right, and so will every field after it. Maybe we could make a different format function specific to history.

ido50:
I put the view date on the far right, so no existing column is pushed, unless I've mistaken your meaning. I started by putting it on the far left (i.e. the first column), but that would have broken backwards compatibility.

With regards to trimming by number of lines, that's actually what I'm doing now, but it's not really what I want. Heavy view days quickly destroy recent activity. I like having a set amount of time to go back to videos I watched.

Let me know how to proceed. Of course I'll re-open the PR on top of the development branch as necessary.

pystardust:
What if we replace the upload date of the video with the date viewed?
Cause the upload date is relative, and having it in the history makes it not very usefull. Implementing it this way would also make it such that the user can see those dates.

ido50:
I could do that. Ideally history would have both in absolute dates, but if the upload date is relative it is indeed not useful in the history. Should I re-implement like this?

pystardust:
Unfortunately, there doesn't seem to be any absolute date in the search result page. Yeah this was seems better

PR #175, opened 2021-03-28
Improve Readme

JojiiOfficial:
Running curl as sudo isn't a good idea if it can be prevented.


pystardust:
@JojiiOfficial also add your pacman repo

PR #181, opened 2021-03-30
Subscriptions straight from Big Google itself

ShrekhavingabadDay:
There is a way of downloading all the subscriptions that belong to a user straight from youtube as a json file.

I added a small function which reads the contents of the "subscriptions.json" (default name when downloaded) file and makes it readable for ytfzf (creates the subscriptions file basically).

Thank you for reviewing my request!

Have a great day!

P.S.: This is my first pull request and I've noticed some smaller issues but I am not sure how to fix them.

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -1104,7 +1104,8 @@ while true; do
 
 	#if -s was specified make another search query
 	if [ $search_again -eq 1 ]; then
-		search_query="" get_search_query
+		search_query=""
+		get_search_query
 		scrape_yt "$search_query"
 	fi
 done
--- a/ytfzf
+++ b/ytfzf
@@ -146,7 +146,7 @@ create_subs () {
 
     for i in $( seq 0 $((sublength-1)) ); do
 
-        channelId=$(cat subscriptions.json | jq --argjson index ${i} '.[$index] .snippet .resourceId .channelId' )
+        channelId=$(cat "$FILE" | jq --argjson index ${i} '.[$index] .snippet .resourceId .channelId' )
         echo "https://www.youtube.com/channel/${channelId:1:24}/videos" >> "$subscriptions_file"
 
     done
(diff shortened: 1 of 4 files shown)


ytfzf
```diff
@@ -126,6 +126,31 @@ dep_ck "jq" "youtube-dl";
...
+    # start fresh
+    > $config_dir/subscriptions
+
+    # check how many subscriptions there are in the file
+    sublength=$(cat "$FILE" | jq '. | length')
+
+    for i in $( seq 0 $((sublength-1)) ); do
+
+        channelId=$(cat subscriptions.json | jq --argjson index ${i} '.[$index] .snippet .resourceId .channelId' )
+        echo "https://www.youtube.com/channel/${channelId:1:24}/videos" >> "$subscriptions_file"
```
pystardust, line 150: I don't think you could do this in posix, we could  maybe you could do something like `${channelId%${channelId#??????????????????????}}`

ShrekhavingabadDay: That didn't do the trick for me, the first double quote wasn't cut off. I piped the variable into 'cut' : 
`$(echo ${channelId} | cut -c 2-25)`

ytfzf
```diff
@@ -126,6 +126,31 @@ dep_ck "jq" "youtube-dl";
...
+   
+    # start fresh
+    > $config_dir/subscriptions
+
+    # check how many subscriptions there are in the file
+    sublength=$(cat "$FILE" | jq '. | length')
+
+    for i in $( seq 0 $((sublength-1)) ); do
+
+        channelId=$(cat subscriptions.json | jq --argjson index ${i} '.[$index] .snippet .resourceId .channelId' )
```
ShrekhavingabadDay, line 149: Whoops... Right here subscriptions.json should be _"$FILE"_.

Issue #183, opened 2021-04-01
ERROR[#01]: Couldn't curl website. Please check your network and try again.


aris-sk:
Greetings...!!! I love this script
But unfortunately since yesterday 31/3/2021 i get this error:

ERROR[#01]: Couldn't curl website. Please check your network and try again.

My network connection is totally fine.
I suspect it 's a change in youtube api or something cause i have problem sending youtube urls to vlc also. Mpv works like a charm though.
Anyone else with same problem?
These are the versions of depedencies i have:
ytfzf: 1.1.1
youtube-dl: 2021.02.04.1
fzf: 0.25.1 (6654239)
jq: jq-1.5-1-a5b5cbe

I would appreciate some help

smtnsk:
Same here. I ran the curl command in the yt_html function with the verbose option and got the following response:

```
*   Trying 142.250.179.206:443...
* Connected to www.youtube.com (142.250.179.206) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN, server accepted to use h2
* Server certificate:
*  subject: C=US; ST=California; L=Mountain View; O=Google LLC; CN=*.google.com
*  start date: Mar 11 14:54:07 2021 GMT
*  expire date: Jun  3 14:54:06 2021 GMT
*  subjectAltName: host "www.youtube.com" matched cert's "*.youtube.com"
*  issuer: C=US; O=Google Trust Services; CN=GTS CA 1O1
*  SSL certificate verify ok.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
[...]

PR #186, opened 2021-04-01
Add option to use wget instead of curl in order to bypass YouTube's consent form

smtnsk:
This patch adds the option to use wget instead of curl in the get_yt_html function. Included are both short and long command line options and the option to set the use_wget variable via conf.sh. The variable defaults to 0, i.e. disabled.

The --no-cookie option of wget allows the user to bypass YouTube's consent form which blocks some users from scraping the site (see #183).

My editor also apparently decided to clean up some trailing whitespace.

### DIFF
--- a/docs/conf.sh
+++ b/docs/conf.sh
@@ -127,6 +127,10 @@ fancy_subscriptions_menu=1
 #yt_subs is the same as -S
 scrape="yt_search"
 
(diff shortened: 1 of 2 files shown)


smtnsk:
Shoot, wrong branch. I'll switch to development.

PR #187, opened 2021-04-01
Add option to use wget instead of curl in order to bypass YouTube's consent form

smtnsk:
This patch adds the option to use wget instead of curl in the get_yt_html function. Included are both short and long command line options and the option to set the use_wget variable via conf.sh. The variable defaults to 0, i.e. disabled.

The --no-cookie option of wget allows the user to bypass YouTube's consent form which blocks some users from scraping the site (see #183).

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -539,14 +539,9 @@ get_yt_json () {
 }
 
 get_yt_html () {
-    link="$1"
-    query="$2"
-
-
-get_yt_html () {
     local link="$1"
     local query="$2"
-	if [ $use_wget -eq 1 ]; then
+	if [ "$use_wget" -eq 1 ]; then
 		printf "%s" "$(
 		wget -O - --no-cookie --quiet \
 	      --header 'authority: www.youtube.com' \
@@ -1038,7 +1033,7 @@ parse_long_opt () {
 
 		notification) parse_opt "N" ;;
 
-		wget) uset_wget=1 ;;
+		wget) use_wget=1 ;;
 
 		version)
 		    printf "\033[1mytfzf:\033[0m %s\n" "$YTFZF_VERSION"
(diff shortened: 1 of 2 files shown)


user122983:
still having issues using this patch
 
ERROR[#01]: Couldn't download website. Please check your network and try again.
Check for new versions and report at: https://github.com/pystardust/ytfzf

is there anything to prevent this

smtnsk:
Are you using BusyBox wget by chance? It lacks the necessary --header and --no-cookies options.

user122983:
arch pacman wget-1.21.1-1

ytfzf
```diff
@@ -537,17 +539,27 @@ get_yt_json () {
...
+    local link="$1"
+    local query="$2"
+	if [ "$use_wget" -eq 1 ]; then
+		printf "%s" "$(
+		wget -O - --no-cookie --quiet \
+		  --header 'authority: www.youtube.com' \
+		  --header "user-agent: $useragent" \
+		  --header 'accept-language: en-US,en;q=0.9' \
+		  "${link}?search_query=${query}&sp=${sp}" \
+		)"
```
pystardust, line 551: why printf the stdout?
We could omit printf since wget will output to stdout.
Similarly with the curl below

pystardust:
Unfortunately I am not able to reproduce this issue in my country to test it out. I would like opinions from others facing this issue regarding this patch.

What if we try the original curl command with `-L` flag?

nigel1901:
Experienced the #183 --header response / curl issue. 
"Learnt" how to get dig thru  "/usr/bin"  and modify the "get_yt_html ()" around line 505 to wget  BUT still had failures. 
OOPS schoolboy error!

Thru copy & pasting the code block from on screen left loads of **'invisible' spaces** after the (command wrapping \) backslashes. 
As soon as these deleted and cleaned up script runs and gets a list of scraped results. 
(Now onto codec problems)

pystardust:
#118 seems to have solved the issue. If it seems to be working well we can close this PR.

smtnsk:
#188 has indeed fixed the issue for me. No need to switch to wget after all, so I'll go ahead and close this.

Issue #190, opened 2021-04-04
[enhancement] add preferred subtitles and add a sorting method for subscriptions


aimad10:
Hello, I was trying to add the following features but I am terrible at programming and like using keyboard shortcuts. 

1. have a config setting to select preferred subtitles. Right now, I have to add --subt to my query and then choose the video, and then choose the language. It would be nice if I could just have it auto select my preferred language (english) and then just not have subtitles if they aren't available instead of having to choose the same thing from the menu everytime I choose a video. 
2. add a sorting method for the subscriptions. Right now it seems that if I do ytfzf -S, it lists the different channels in a random? order, and then lists the videos that the channel posted in chronological order. It would be nice if it were to display the channels in chronological order based on the most recent video, so if a channel posts, it get bumped up to the top of the order.

How it is right now:
---CHANNEL A--- 
video1 -- 2 min ago
video2 -- 2 weeks ago
video3 -- 1 year ago

---CHANNEL B---
video1 -- 1 week ago
video2 -- 2 months ago
video3 -- 7 years ago

---CHANNEL C---
video1 -- six seconds ago
video2 -- 2 min ago
video3 -- 3 weeks ago

How I propose it should be:

---CHANNEL C---
video1 -- six seconds ago
video2 -- 2 min ago
video3 -- 3 weeks ago

---CHANNEL A--- 
video1 -- 2 min ago
video2 -- 2 weeks ago
video3 -- 1 year ago

---CHANNEL B---
video1 -- 1 week ago
video2 -- 2 months ago
video3 -- 7 years ago

Not sure if this is the right place to do this but really, thanks for making this. I use linux inside of a VM in windows so I can play games and use linux at the same time, and this tool fixed adblock and the framskipping issues at once. Really glad you made this.

westofer:
for number 1
there are few options


## Using "YTFZF_SELECTED_SUB", I didn't test it, it is not stable yet

look at pull request #193 for details


## Using mpv config

add to `~/.config/mpv/mpv.conf` the following line

    slang="your-language-code,second-prefered langauge"
    
    # Example, use english, if not avialble use japanese
    slang="en,jpn"

## Using YTFZF vars

change mpv options ONLY when used with ytfzf

in `~/.config/ytfzf/conf.sh` add the following line

    YTFZF_PLAYER="mpv --slang=en,jpn"

spector-9:
You have to explain how is it unstable. If you encountered a bug/s open a new issue with details.

[...]

PR #193, opened 2021-04-05
  Added an option to set a default language for auto gen captions

spector-9:
Requested in #190 first part. You can set the default language with variable "YTFZF_SELECTED_SUB". Also I changed to method of getting sub_list variable to original because the new method does not differentiate between auto generated subs and manual ones. Which causes errors because method is designed for auto gen subs. You can check the same by playing a video with manual subs.


westofer:
can documentaion be add to `--help-all` please

btw `YTFZF_PLAYER` and `YTFZF_PLAYER_FORMAT` aren't documentated too

PR #202, opened 2021-04-09
Last search

Euro20179:
allows for `!!` and `![number]` searches (helpful when -s is used)


pystardust:
It would be better if we cache those results for a session instead of scarping again.

Issue #203, opened 2021-04-09
Audio only not working


qoheniac:
When I use the `-m` flag I get an error message saying the requested file format isn't available and then the video is played instead of just audio. If I use the `-f` flag and choose Audio no error shows up but it still plays the video.
```
$ youtube-dl --version
2021.04.01
$ ytfzf -v 
ytfzf: 1.1.2
$ ytfzf -am low-fi
Scraping Youtube...
Opening Player: mpv
[ytdl_hook] ERROR: requested format not available 
[ytdl_hook] youtube-dl failed: unexpected error occurred 
Failed to recognize file format.

Exiting... (Errors when loading file)
 (+) Video --vid=1 (h264 1920x1080 30.000fps)
 (+) Audio --aid=1 (aac 2ch 48000Hz)
AO: [pulse] 48000Hz stereo 2ch float
VO: [gpu] 1920x1080 yuv420p
AV: 00:00:02 / 00:00:14 (15%) A-V:  0.000 Cache: 12s/1MB

Exiting... (Quit)
```

pystardust:
This seems to be a livestream.
youtube-dl is trying to get the audio. But since its a livestream, youtube-dl is not able to stream `bestaudio` format.  ytfzf falls back to normal format if youtube-dl fails to play video in the specified format. 
I just check the available formats. There doesn't seem to be any audio only version of a livestream using youtube-dl.

qoheniac:
Then how about using `mpv`'s `--no-video` flag for streams? Like this:
```
mpv --no-video 'https://www.youtube.com/watch?v=5qap5aO4i9A'
```

pystardust:
yes we could do that,
As a temporary fix. Parse the env variables
```sh
YTFZF_PREF=  YTFZF_PLAYER="mpv --no-video"  ytfzf -a low-fi
```
Or add it to the conf.sh

qoheniac:
Thanks. Apart from that, I was wondering why it shows an error message for the `-m` flag but not for the `-f` flag after choosing `Audio` and I think it's because in [this line](https://github.com/pystardust/ytfzf/blob/be24919df838aaf4badffab77f77a716be9fcdb0/ytfzf#L1080) we should have `"bestaudio/best"` instead of just `"bestaudio"`.

pystardust:
The PR works great. We can  close this issue if its resolved.

qoheniac:
Sure, thank you for swift feedback! :)

PR #205, opened 2021-04-09
Solve issues with audio-only playback

qoheniac:
See https://github.com/pystardust/ytfzf/issues/203


pystardust:
`--no-video` disables video playback. But if we choose a format with video, won't youtube-dl buffer the video too?

pystardust:
Opps, I just read the manual you quoted, it doesnt download the video

qoheniac:
No problem, I think I wrote that edit just the moment when you wrote your comment here.

pystardust:
Nice, this would work for mpv. It might be an issue for external players likes vlc
We would need to take care of them too

pystardust:
Send the PR to development we could merge and work on it there

qoheniac:
Here you go: https://github.com/pystardust/ytfzf/pull/206

qoheniac:
> It might be an issue for external players likes vlc

I wanted to look into this but how do I even use other players? If I simply try `YTFZF_PLAYER=vlc ytfzf something` with version 1.1.2 and pick some video I always get this error message: `http stream error: local stream 1 error: Cancellation (0x8)`.

PR #211, opened 2021-04-12
Better add subs

Euro20179:
adds a comment that says the channel name for each sub in subscriptions.json in addition, all spaces need to be removed from subscriptions.json or urls will break


ytfzf
```diff
@@ -941,8 +942,7 @@ scrape_subscriptions () {
 	done << EOF
 $( sed \
 	-e "s/#.*//" \
-	-e "/^[[:space:]]*$/d" \
-	-e "s/[[:space:]]*$//" \
+	-e "s/[[:space:]]*//g" \
```
pystardust, line 945: we could remove the lines which are empty
`/^[[:space:]]*$/d`
else the while read loop will take an empty line

pystardust: or maybe add a check to see if $url in empty before scrape_channel

ytfzf
```diff
@@ -159,10 +159,11 @@ create_subs () {
     # check how many subscriptions there are in the file
     sublength=$( jq '. | length' < "$yt_sub_import_file" )
 
+    subscriptions_json="$(cat "$yt_sub_import_file")"
```
pystardust, line 162: the way it was before is better. Since we are calling lesser subshells. Storing in a variable means we would need to pipe it, we would be invoking another subshell due to the pipe

PR #212, opened 2021-04-15
Newer than

Euro20179:
this is extremely useful for subscriptions because some channels upload once a week/month or every few months and it's nice to not see their old videos every time you do `ytfzf -S`


pystardust:
Instead of getting newer than and older than we could first implement sort in the similar way. That we could truncate that sorted video list after the older/newer dates.

Euro20179:
i have tried implementing a proper sort and could not figure out how

PR #219, opened 2021-04-17
Add notify-send and shuf to the optional dependencies

### LINKED ISSUE #216: Documentation should mention shuf as a required dependency
As I am porting ytfzf to FreeBSD, I've spotted that it requires shuf to be present on the system. The shuf utility is not one of the portable utilizes you can expect your UNIX-like system to ship with by default.

It would be cool if the documentation mentioned that ytfzf requires shuf (I am not sure if there are any other utilities like that being used).

Cheers!

### LINKED ISSUE #217: Documentation does not mention libnotify as a dependency
Ytfzf seems to be using `notify-send` which is a part of the `libnotify` package.

0mp:
Fixes #216 and #217.


0mp:
Simultaneously, we should handle gracefully the case when those two utilities are missing. Currently, there are no checks AFAIR.

PR #225, opened 2021-04-21
better options, way to do something in config when option is parsed

Euro20179:
i use a very crappy way of testing when -S is passed in my .config
```sh
if [ -n "$(printf "%s" "$1" | grep "\-S")" ]; ...```
this would allow for a simple function

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -532,7 +532,7 @@ get_sp_filter () {
 	if [ -n "$upload_date_character" ]; then
 		#replaces the 8th character in the filter_id with the appropriate character
 		#the 8th character specifies the upload_date_filter
-		sp=$(printf "%s" "$filter_id" | sed "s/\\(.\\{7\\}\\)./\\1$upload_date_character/")
+		sp=$(printf "%s" "$filter_id" | sed 's/\(.\{7\}\)./\1'"$upload_date_character"'/')
 	#otherwise set it to the filter_id
 	else
 		sp=$filter_id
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by pystardust** [comment]
I understand the point. But we already have all the variables we can set in the conf.sh.
This might be a nice addition but kinda overkill. What do you think?
Also the eval statement in the parse opt might mess up string passed in as arguments. something like `ytfzf "A1 A2" "B1 B2" ` Would become `ytfzf A1 A2 B1 B2`

**Comment by Euro20179** [comment]
> Also the eval statement in the parse opt might mess up string passed in as arguments. something like function "A1 A2" "B1 B2" Would become function A1 A2 B1 B2

maybe run the eval statement after the opt is parsed

**Comment by pystardust** [comment]
We could try maybe using the `set` builtin with the config opts  inside a function and running parse opt again after the normal opts are parsed.

**Comment by Euro20179** [comment]
can you write out what you mean?

File: ytfzf [hunk 1]
```diff
@@ -938,7 +938,7 @@ update_ytfzf () {
 	updatefile="/tmp/ytfzf-update"
 	curl -L "https://raw.githubusercontent.com/pystardust/ytfzf/$branch/ytfzf" -o "$updatefile"
 
-	if sed -n '1p' < "$updatefile" | grep -q '#!/bin/sh' ; then
+	if [ -n "$(cat "$updatefile")" ];  then
```
pystardust, line 941: If there was a curl error or gh servers are down, the we get a 404 response. 
so the file won't be empty, it will have some error statement. Thats why I was checking for the shebang

Euro20179: ok

PR #228, opened 2021-04-24
Add numbers next to queued videos

TomBT:
Add numbers next to queued videos by order in which they were queued. Say the user queues up videos, switches to another desktop, switches back to the queued videos and forgets the order of the queue. Numbers next to the videos would make that order apparent. I made a crude [photoshop](https://raw.githubusercontent.com/TomBT/imge/main/2021-04-23-195026_1366x768_scrot.png) of what my suggestion might look like


TomBT:
OOPS! I randomly queued the videos so don't look at number 5. 😳

PR #230, opened 2021-04-26
Last search

Euro20179:


test
```diff
@@ -0,0 +1 @@
+
```
pystardust, line 1: remove new file

ytfzf
```diff
@@ -704,6 +708,21 @@ scrape_yt () {
...
+parse_special_search () {
+    search="$1"
+    number="$(printf "%s" "$search" | grep -Po "[\-0-9]+")"
+    if [ "$search" = "!!" ]; then
+	number="$search_count"
+    elif [ "$number" -lt 0 ]; then
+	#needs to be + because number is negative
+	number="$((search_count + number + 1))"
+    fi
+    if ! cat "${tmp_searches_prefix}${number}"; then
```
pystardust, line 720: to check if file exists you could use the text expression like so
```sh
[ -f $filename ]
```

ytfzf
```diff
@@ -720,13 +739,23 @@ get_search_query () {
 			printf "$search_prompt"
 			read -r search_query
 		fi
-		[ -z "$search_query" ] && exit 0
+		if [ -n "$(printf "%s" "$search_query" | grep -Po '!(?:-?[0-9]+|!)')" ]; then
```
pystardust, line 742: look at the `-q` option of grep. It will do the operation quietly and return only exit status on match/non-match. So you wouldn't need to check if non empty and directly do something like
```sh
#!/bin/sh
if echo "hi" | grep -q "hi" ; then
	echo "do stuff.."
fi

```

ytfzf
```diff
@@ -704,6 +708,21 @@ scrape_yt () {
 	unset videos_json yt_json yt_html
 }
 
+parse_special_search () {
+    search="$1"
+    number="$(printf "%s" "$search" | grep -Po "[\-0-9]+")"
```
pystardust, line 713: since we already know that the string satisfies the regex is it possible to trim of the `!` to get the number? Maybe with a parameter expansion with `${var#!}`

ytfzf
Euro20179: this is like this because if cat doesn't return an error, the contents of the file is printed and is returned to where the function is called, otherwise it returns invalid, but using `[ -f $filename ]` it looks like this

```sh
if [ ! -f "${tmp_searches_prefix}${number}" ]; then
    printf "%s" "$search is an invalid lookback"
    clean_up
    exit 2
else
    cat "${tmp_searches_prefix}${number}"
fi```

it's just more explicit

**Comment by Euro20179** [comment]
dont merge yet

**Comment by Euro20179** [comment]
ok, now merge if it looks good, i added a way to be able to use a lookback that is !search

**Comment by pystardust** [comment]
```sh
./ytfzf -s search1                                                           
Scraping Youtube...
Opening Player: mpv https://www.youtube.com/watch?v=YUH7QYl6qQo

Exiting... (Quit)
Search Youtube: search2
Scraping Youtube...
Opening Player: mpv https://www.youtube.com/watch?v=5pEX-naQ-7c

Exiting... (Quit)
Search Youtube: !!
```
This opens fzf with invalid loopback.
```
shell: zsh
terminal emulator: st
```

It doesn't clear up the tempfiles and uses them again if the exit was abrupt, I was wondering if we could use only file and store all searches in there. 
Globbing a file name seems a less secure. 
Maybe we could use the cache dir instead.

Euro20179:
> A minimal approach to this could be to use only one file to store old searches. And the user could fuzzy find what they wanted from all the previous searches

this accomplishes what i was trying to do, which is make the user not retype a search, so yeah sure

PR #232, opened 2021-04-29
remove unecessary check

Euro20179:


ytfzf
```diff
@@ -1209,8 +1209,10 @@ parse_opt () {
 
 
 while getopts "LhDmdfxHaArltSsvNTn:U:-:" OPT; do
-	# to parse extra-options in conf.sh
-	function_exists "on_opt_parse" && on_opt_parse "$OPT" "$OPTARG"
+	# to parse extra-options in conf.sh ( and don't parse during thumbnail call)
+	if function_exists "on_opt_parse"; then
+	    [ "$OPT" = "-" ] && on_opt_parse "${OPTARG%%=*}" "${OPTARG#*=}" || on_opt_parse "$OPT" "$OPTARG"
```
pystardust, line 1214: This is nice that we can parse long opts directly from here. 
Should we consider doing this to the regular `parse_opt`?
That might cut down a lot of code

pystardust (approved):

ytfzf
Euro20179: i can try it out

PR #234, opened 2021-05-01
[refactor]: change how env + environment variables are set and used

Euro20179:
this also fixes a bug where in on_opt_parse, when setting an environment variable you can't use the config name because it doesn't get read.


ytfzf
```diff
@@ -17,60 +17,58 @@ config_file=${YTFZF_CONFIG_FILE-$config_dir/conf.sh}
 tmp_video_data_file="/tmp/ytfzf-subdata"
 #source config file if exists
 #it also checks to make sure that this is not a subprocess of another ytfzf script
-[ -e "$config_file" ] && [ -z "$PROC_ID" ] && . "$config_file"
+[ -e "$config_file" ] && . "$config_file"
```
pystardust, line 20: So this is to make the option `-U` work with `on_opt_parse` right?

Euro20179: the comment and branch name are irrelevant

the problem was that if config variables that could also be used as environment variables were defined in on_opt_parse it wouldnt use the variables because it uses the env version of the variable in the program which get set before on_opt_parse is called and therefore before the variables are set.

pystardust: Ah that makes it clear

PR #247, opened 2021-05-16
refactor: change how opts are parsed

Euro20179:
much, much less repetition.

this pr also adds a feature, where you can disable the error message if you pass in an opt that doesn't get read by ytfzf, this enables the use of opts that are read by on_opt_parse only,

eg: in my config i have a --live opt which filters by live.


pystardust:
Interesting. If someone whats a specific feature then we could write them a config file instead of changing the source. And the code look really neat merging the long opts.

ytfzf
```diff
@@ -1031,186 +1033,148 @@ bad_opt_arg () {
...
 	esac
 	unset opt optarg
 }
 
 
 while getopts "LhDmdfxHaArltSsvNTn:U:-:" OPT; do
+	# to parse extra-options in conf.sh
+	[ "$OPT" = "-" ] && OPT="${OPTARG%%=*}" && OPTARG="${OPTARG#*=}"
+	#when there is no = in OPTARG and it's a longopt, OPTARG will = OPT
+	[ "$OPTARG" = "$OPT" ] && OPTARG=""
```
pystardust, line 1174: Since this part is only when `$OPT` is `-`. It would be a lil tidy to have an `if` statement here, for both these lines

PR #248, opened 2021-05-16
Fixed linting issues

paniash:


pystardust:
I have gone through shell check. Number need not be quoted. And I have not quoted numbers to make it easier to know which variables are numbers and strings. Most `"%s"` missing are also design. Since the variable may have the escape sequence. Like the prompts. The `set` in the sort function would break with the quotes. Since we want word splitting.

PR #250, opened 2021-05-16
feat: event functions

Euro20179:
this pull request adds many event functions that act similarly to on_parse_opt
it also adds customizable sorting methods

## Variables

* `custom_videos_data`
    *  when videos_data is provided in a user defined function this should also be set to 1 so that ytfzf doesn't override videos_data

#### `send_select_video_notif`
gets called when a video is selected
this function handles sending the notification

**usage**:
* custom notifications

#### `on_video_data_gotten`
gets called after `videos_data` is set in `scrape_yt` and `scrape_channel`
there is no expectation of what this function does,

**parameters**:
1. `videos_data`
2. `videos_json`
3. `yt_json`

**usage**:
* to do stuff with videos_data, eg write it to a file,
* make a custom videos_data menu with videos_json/yt_json

#### `on_get_search`
gets called when the search is gotten
there is no expectation of what this function does

**parameters**:
1. `search_query`

**usage**:
* custom actions when a specific search is given
* logging searches to a file

#### `on_exit`
gets called when the last instance of ytfzf exits
there is no expectation of what this function does
**usage**:
* a goodbye notification,
* cleaning up user created tmp files

## Sorting

allow custom sorting methods through 2 functions
`data_sort_key` and `data_sort_fn`

#### `data_sort_key`
gets called to get the value to sort by and to put in `data_sort_fn`

**parameters**:
1. video_title
2. video_channel
3. video_length
4. video_views
5. video_date
6. video_shorturl

#### `data_sort_fn`
gets called to sort the data

### Sort names
a sort name is a function that defines `data_sort_key` and `data_sort_fn`
eg:
```sh
alpha () {
    data_sort_key () {
        sort_by="${1#|}"
        printf "%s" "$sort_by"
    }
    data_sort_fn () ){
        sort
    }
}
```

to use a sort name, in config define the variable `sort_name=alpha` or with the command `ytfzf --sort-name=alpha`

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -880,9 +880,9 @@ session_is_running () {
 	[ -d /proc/"$pid" ] && session_count=$(( session_count + 1 ))
     done < "$pid_file"
     if [ $session_count -eq 1 ]; then
-	return 0
-    else 
 	return 1
+    else 
+	return 0
     fi
 }
 #> removes tmp files and clutter
@@ -895,7 +895,6 @@ clean_up () {
 }
 #> Saves data before exiting
 save_before_exit () {
-	clean_up
 	[ $is_url -eq 1 ] && exit
 	[ $enable_hist -eq 1 ] && printf "%s\n" "$selected_data" >> "$history_file" ;
 	[ $enable_cur -eq 1 ] && printf "" > "$current_file" ;
--- a/ytfzf
+++ b/ytfzf
@@ -1016,7 +1016,6 @@ sort_video_data_fn () {
 	while IFS= read -r line
 	do
 		IFS="$tab_space"
-		set -- $line
 		#run the key function to get the value to sort by
 		printf "%s\t%s\n" "$(data_sort_key $line)" "$line"
 	done | data_sort_fn | cut -f2-
--- a/ytfzf
+++ b/ytfzf
@@ -77,6 +77,8 @@ search_prompt=${search_prompt-Search Youtube: }
 useragent=${useragent-'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.152 Safari/537.36'}
 #whether or not to show errors when an invalid option is passed
 exit_on_opt_error=${exit_on_opt_error-1}
+#whether or not you provided videos_data
+custom_videos_data=0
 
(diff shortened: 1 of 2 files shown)


ytfzf
```diff
@@ -620,10 +627,14 @@ scrape_channel () {
 	]')
 
 	videos_json=$(printf "%s" "$videos_json" | jq '.[0:'$sub_link_count']')
-	videos_data=$(get_video_data "$videos_json")
+	[ -z "$videos_data" ] && videos_data=$(get_video_data "$videos_json")
```
pystardust, line 630: Here, why do we need to check for `videos_data` being empty?

Euro20179: `videos_data` gets reset every loop when there is a new search, checking if its empty allows for people to set videos_data in on_get_search or any function set in config without it being overwritten. Doing this allows an implantation of !! searches purely in config

ytfzf
```diff
@@ -1217,6 +1275,8 @@ check_if_url "${search_query:=$*}"
 #format the menu screen
 format_menu
 
+[ -n "$sort_name" ] && $sort_name
+unset sort_name
```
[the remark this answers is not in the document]
Euro20179: I guess the check is unecessary because If its unset nothing happens

unset was used everywhere else so I just assumed I could use it but the script should work fine without it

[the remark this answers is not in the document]
pystardust: my bad thats right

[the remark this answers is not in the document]
Euro20179: well i've removed it so if there's a need to add it back then we can add it back, but if you wanted to check what sort_name was used you have the option when it's not unset

ytfzf
```diff
@@ -954,16 +996,31 @@ update_ytfzf () {
...
+	while IFS= read -r line
 	do
 		IFS="$tab_space"
-		set -- $_VLine
-		_Date=${5#|}
-		_Date=${_Date#Streamed}
-		printf "%d\t%s\n" "$(date -d "${_Date}" '+%s')" "$_VLine"
-	done | sort -nr | cut -f2-
-	unset IFS _VLine _Date
+		set -- $line
```
pystardust, line 1019: since we are using `IFS` into a function `data_sort_key`, this `set` will not be getting used

Euro20179: removed

Issue #44, opened 2021-02-27
MacOS imag-preview


Jac-Zac:
Just to make you aware of this, preview of thumbnails doesn't work on macOS. The script download the images were it should, but it doesn't show them. It looks like the problem has to do with Überzug. I also tried to run fontpreview-ueberzug which won't work either. But I might also be wrong so if other people have mac they might be helpful. I also tried to used Überzug to preview images in ranger and It doesn't work. In ranger they use a preview for that works with iTerm terminal emulator. An example is imgcat

westofer:
can you post the output of `ueberzug version`?

mine is `18.1.9`

Jac-Zac:
> ueberzug version

Same... Does it work for you ?

westofer:
yes it works, are you using fish by any chance? becuase it doesn't work in fish for me

Jac-Zac:
> yes it works, are you using fish by any chance? becuase it doesn't work in fish for me
> 
> @Jac-Zac

I’m using zsh maybe we can have a discord server and talk there instead of just messaging ?

westofer:
sure @Jac-Zac

Jac-Zac:
https://discord.gg/FNym2zpK This is the discord server if we find helpful we might also expand it in the future and add it to README.md

Z5483:
Ueberzug relied on X11 so it wouldn't work in system that doesn't use X11

I know it's late but this might be helpful to some people who have this issue

pystardust:
Yes people on wayland are also having this issue. If anybody knows a terminal viewer that work on wayland + mac do let us know.

cybarspace:
> Yes people on wayland are also having this issue. If anybody knows a terminal viewer that work on wayland + mac do let us know.

Would w3m work?

tinywrkb:
@pystardust Ranger file manager has iTerm2 support for image previews, so maybe look at what it's doing. I don't have a Mac but it worked for me with Alacritty when pulling in [the graphics PR](https://github.com/alacritty/alacritty/pull/4763) that adds iTerm2 protocol & libsixel (though it looks like it won't be merged as it is).  
Ranger also has support for the image protocol of Kitty, and Kitty is also cross-platform.

Z5483:
I also like the way `neofetch` provides different image preview options for the user to choose from. I understand it would take a long time to implement, but it would be a neat and cool feature.

dimitarvp:
[...]

PR #254, opened 2021-05-19
feat: more ways to view thumbnails

Euro20179:
addresses #44


piotryordanov:
Can it be made to support `w3m`?

Euro20179:
It would be nice to be able to support w3m, however we have not found a way to get w3m to work with fzf

piotryordanov:
Oh okay. So, is chaca the only alternative option for wayland?

Euro20179:
currently yes, sadly, so far the only things that work are programs that print ascii characters and color them.

piotryordanov:
THank you for answering.
So, is that an fzf limitation? Or, a limitation of FZF running in Wayland?

Euro20179:
To my knowledge it's either a limitation of fzf or curses (the library that fzf uses)

PR #257, opened 2021-05-23
more efficient ifs

Euro20179:
add-playlist got changed to playlist-add because it doesn't add a playlist, it adds to a playlist (also in help it displayed playlist-add)

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -818,7 +818,7 @@ get_search_query () {
 			printf "$search_prompt"
 			read -r search_query
 		fi
-		[ -z "$search_query" ] && (clean_up; exit 0)
+		[ -z "$search_query" ] && clean_up && exit 0
 	fi
 	function_exists "on_get_search" && on_get_search "$search_query"
 }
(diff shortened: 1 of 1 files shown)


ytfzf
```diff
@@ -818,7 +818,7 @@ get_search_query () {
 			printf "$search_prompt"
 			read -r search_query
 		fi
-		[ -z "$search_query" ] && clean_up && exit 0
+		[ -z "$search_query" ] && (clean_up; exit 0)
```
pystardust, line 821: I don't know if this is intensional but `()` will run the code in a sub shell (won;t make much of a difference here). Use `{}` for grouping commands with out subshell

Euro20179: () makes exit only happen if [ -z $search_query Is true

pystardust: if you do `(exit 0)` it will not exit the script. It will only exit the subshell

Euro20179: fixed

PR #284, opened 2021-06-14
refactor: better looking thumbnail view

pystardust:
the thumbnail view with the description looks very ugly. I moved the image down a little bit and put the description on a new line.
It still need some more touch up i guess


Euro20179:
for the `padding` thing it might be better to let people export that in their `FZF_DEFAULT_OPT` in the config file

Issue #274, opened 2021-06-05
Search again hotkey


p1839251:
Hello,
It would be useful if there were a search again hotkey.
I use `ytfzf` with `-l `(loop) so I could stay inside the program.
I look something up, browse a little, watch a couple of videos then I sometimes want to modify the search terms.
Right now, I have to exit and type `ytfzf` again, which isn't very convenient.

Euro20179:
#289 and update to the development branch with `ytfzf --update-unstable` (alt-s is the hotkey to search again)

PR #288, opened 2021-06-17
New Feature: Add alt-keys to fzf video menu

simonhughcom:
First I wanted to say, great work on ytfzf, I only discovered it yesterday and I am thoroughly enjoying it, and of course I immediately had a few ideas to contribute :D

Adds a series of alt-keys to the fzf video menu.
Allows the user to choose what to do with the selection inside the FZF menu.

These alt-keys have currently been implemented:
    alt-l : Prints the selected URL only
    alt-t : Prints the selected title only
    alt-o : Opens selected URL in BROWSER
    alt-v : Play selected video
    alt-m : Play Audio only (for music)
    alt-d : Download to current directory

I can implement more if you can think of any other ideas.


Euro20179:
This is very cool! I've been looking for a way to implement something like this, I know it solves a couple of bug reports, eg: #274, it is much more elegant than my solution of having an option to select. (in fact I am going to remove it)

I have re implemented this, the shortcuts you added are the same and will work the same, and it still uses fzf's `--expect` parameter. What I added is a way to add more shortcuts yourself, and to be able to change the default shortcut bindings in the config file.
I also mentioned the shortcuts in docs/USAGE.md.
I have also renamed the function `alt_keys` to `handle_shortcuts`

When exiting, ytfzf needs to do some clean up with the clean_up function, so instead of exiting directly in the function, it will return 1 when it wants to exit and 0 when it doesn't, that way ytfzf can clean up.

I will be making a new pull request (#289), you can review it and if something seems off tell me, Im not sure what setsid does, but from what i gathered it just forks a command and makes it it's own process which can be achieved like this `command & disown`

In addition, you seemed to have created a branch from the master branch and not the development branch, this just creates a lot of commits and merge conflicts, and you should try to avoid it.

Hopefully this comment wasn't too negative, this is a very cool pull request!

PR #289, opened 2021-06-17
feat: shortcuts (credit to @simonhughcom (#288)) 

Euro20179:
shortcuts in fzf, finally

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -73,6 +73,20 @@ If you wish to find inspiration and see an example you can take a look at the [`
 	+ `-m`: (audio-only) Only audio format (best audio)
 
 	+ `-d`: (download) To **download** a video instead of playing it.
+	
++ ### Fzf Shortcuts
+
+    These shortcuts will currently only work in fzf
+
+    + `alt-l`: print the selected urls
+    + `alt-t`: print the selected video(s)' title
+    + `alt-o`: open the selected video(s) in a browser
+    + `alt-v`: watch the selected video(s)
+    + `alt-d`: download the selected video(s)
+    + `alt-m`: listen to the audio of the selected video(s)
+    + `alt-s`: search again
+
+    These shortcuts can all be changed in `~/.config/ytfzf/conf.sh` (see the [config](conf.sh) file and go to the `MISC` section for more)
 
 + ### Multi-selection
 
--- a/docs/conf.sh
+++ b/docs/conf.sh
@@ -148,13 +148,6 @@ sub_link_count=10
 #same as --fancy-subs=
 fancy_subscriptions_menu=1
 
-#whether or not to show the search again button
-#1 means it's  always on
-#0 means always off
-#empty means enable when search_again and enable_loop are on
-#same as -ls
-search_again_button=
-
 #where to source videos from
 #options are: history, yt_subs, yt_search, trending, pt_search
 #history is the same as -H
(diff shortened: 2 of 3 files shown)


simonhughcom:
All looks good, except for `disown` which is not posix, its not available is dash for example.
You can also use `nohup` which is posix compliant.

Euro20179:
Thanks, fixed.

pystardust:
Really nice implementation

consider the following refactors:

* Title may contain pipe
```diff
-       "$title_shortcut") printf "%s\n" "$selected_data" | cut -d '|' -f1; return 1 ;;
+       "$title_shortcut") printf "%s\n" "$selected_data" | cut -d "$tab_space" -f1; return 1 ;;
```
* unneeded sed
```diff
        #through this loop, the selected data which was truncated by formatting is retrived.
-       selected_key=$(printf "%s" "$selected_data" | sed 1q)
+       selected_key=${selected_data%%${new_line}*}
        selected_urls=
```
* formatting huge lines with breaks
```diff
-    IFS="," read -r urls_shortcut title_shortcut open_browser_shortcut watch_shortcut download_shortcut audio_shortcut search_shortcut << EOF
-$shortcuts
-EOF
-    case $selected_key in
+       IFS="," read -r \
+               urls_shortcut title_shortcut open_browser_shortcut watch_shortcut \
+               download_shortcut audio_shortcut search_shortcut <<-EOF
+       $shortcuts
+       EOF
```

Euro20179:
Everything is good except the first one, the title shortcut for some reason only works with thumbnails, otherwise it displays the entire selection. which is why i split by pipe, but you are correct, so i'll look for another way.

pystardust:
Ah yes, while running without thumbnails we are using the column command with out the output separator as its not posix. So we don't have the tab separation on the data. Maybe we could stick with the pipe for now

Euro20179:
> Maybe we could stick with the pipe for now

I actually think this is a bad idea because a lot of videos have a pipe in them.

pystardust:
we would need to grep the line out from the data using the surl , that would need an additional function. We can't do much with the raw fzf output

Euro20179:
I have an idea, instead of seperating by tab, we could separate by the length of the titlle gap. I just don't know how to do that yet.

pystardust:
I think the length is determined by the column command. We tried to manually set the length  using awk's printf, but that failed on emoji. So we stuck with `column`.
What if we suffix the title with `\t` in jq itself?

Euro20179:
I tried this, I believe it doesn't work becase the gap size would be longer than than the actual gap size, because the title eats up more of the padding.
```sh
gap="$(printf "%-${title_len}.${title_len}s\t" "t" | sed 's/t//')"
```

pystardust:
even if the title were small, this wouldn't work since column is using tabspaces as its separator

Euro20179:
i actually tested this myself to be sure, a title on youtube can't have more than 1 space in it, I'm going to separate by 2 spaces using awk, because cut can't use more than 1 character as a separator. Im not positive about peertube but i doubt they allow more than 1 space.

Euro20179:
ok, i've applied your changes.

pystardust:
> i actually tested this myself to be sure, a title on youtube can't have more than 1 space in it, I'm going to separate by 2 spaces using awk, because cut can't use more than 1 character as a separator. Im not positive about peertube but i doubt they allow more than 1 space.

This feature won't be a really crucial feature as it outputs only the name of the video, so I think is fine

Seems good lets merge it. Maybe we can test it out and push to master

pystardust (approved):

PR #291, opened 2021-06-18
feat: --detach the video player

pystardust:
* using setsid -f


mrlyht:
Great! I used to homebrew this behavior in a much less elegant way....
Had to reaally dig to find how to use it, though. It should be added to the using documentation.

Also, I really like this behavior so I set the control variable to 1 in the config file, worked fine until I tried the --download flag. It only played, no download. Could you adjust it? Or at least teach me the ropes so I can fix it?

Thank you!!

Euro20179:
#316 (you can either clone the development branch manually or run `ytfzf --update-unstable` to get this update)

PR #292, opened 2021-06-18
feat: Add subscribe fifo

simonhughcom:
This is not finished yet, I just wanted to get your feedback and start a discussion.

I got this idea from bspwm, the idea is to expose the user selection through a named pipe to allow the user to write his own scripts with the selection while keeping the menu persistent.

The user would launch `ytfzf -l -X` (-X was one of the few available flags), and then use `ytfzf --sub` to get the user selections.

the output of `ytfzf --sub` would be along the lines of `COMMAND: DATA`, for example `DOWNLOAD: http://...`

At the moment the `alt-s` going back to search does not work.

Edit: It might be better that instead of having `COMMAND: DATA` we could have `KEY: DATA` and expose the keys the user pressed, for example `alt-o: https://...`, this would reduce code complexity and give the user more control.

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -307,6 +307,66 @@ print_error () {
     [ $ext_menu_notifs -eq 1 ] && notify-send "Check for new versions and report at: https://github.com/pystardust/ytfzf\n" || printf "Check for new versions and report at: https://github.com/pystardust/ytfzf\n" >&2
 }
 
+
+############################
+#        Subscribe         #
+############################
+
+SUBSCRIBE_FIFO=/tmp/ytfzf.fifo
+
+mk_subscribe(){
+    [ -e $SUBSCRIBE_FIFO ] || { mkfifo "$SUBSCRIBE_FIFO" || exit 1 ; }
+}
+
+update_subscribe(){
+    if [ -e $SUBSCRIBE_FIFO ]; then
+        printf "%s\n" "$1" >> $SUBSCRIBE_FIFO
+    fi
+}
+
+
+handle_subscribe () {
+    if [ -z "$selected_key" ]; then
+        printf "VIDEO: %s\n" "$selected_urls"
+        return 1
+    fi
+    #creates splits the variable shortcuts by , and assigns a variable to each string
+    #ei: $urls_shortcut will be equal to the first keybind
+    IFS="," read -r \
+	urls_shortcut title_shortcut open_browser_shortcut watch_shortcut \
+	download_shortcut audio_shortcut search_shortcut <<-EOF
+$shortcuts
+EOF
+    case $selected_key in
+	"$urls_shortcut") printf "URL: %s\n" $selected_urls; return 1 ;;
+	"$title_shortcut") 
+	    printf "TITLE: %s\n" "$selected_data" | awk -F "  " '{print $1}'; return 1 ;;
+	"$open_browser_shortcut") printf "BROSWER: %s\n" "$selected_urls"; return 1 ;;
+	"$watch_shortcut") printf "VIDEO: %s\n" $selected_urls; return 1 ;;
+	"$download_shortcut") printf "DOWNLOAD: %s\n" $selected_urls; return 1 ;;
+	"$audio_shortcut") printf "AUDIO: %s\n" $selected_urls; return 1 ;;
+	"$search_shortcut") #  TODO: Fix going back to search
+	    unset videos_data search_query
+	    [ "$scrape" = "pt_search" ] && scrape_fn || scrape="yt_search" scrape_fn
+	    user_selection
+	    format_user_selection
+	    return 0;;
+    esac
+    function_exists "handle_custom_shortcuts" && handle_custom_shortcuts
+    return $?
+}
+
+subscribe(){
+    if [ -e $SUBSCRIBE_FIFO ]; then
+        tail -f $SUBSCRIBE_FIFO
+        exit
+    else
+        echo "ERROR"
+        exit 1
+    fi
+}
+
+
 ############################
 #        Formatting        #
 ############################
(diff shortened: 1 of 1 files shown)


Euro20179:
my code for alt-s didn't work, take a look at #289 where it returns 2 to signal to restart the loop,
also can you show an example of how one would interact with this in a script, because i don't quite get it.
I think it might be better to instead have a way to overwrite the behavior of the shortcuts in the config file.

pystardust:
Could you explain it again, im not sure if I understood it.
My guess is
If the user wants to do selection with their own scripts, then we can output the selection data into a datafile. And open `fifo` and read for the selected videos there. The user could use any menu to parse that datafile into their selection script, and output the selected options into the fifo. From their the script works as it is taking the input from the fifo as the selected videos.

simonhughcom:
Yes I know, the `alt-s` still needs to be fixed, this is just a proof of concept at the moment.

One would interact with this through `ytfzf --sub`, which just returns the commands to stdout.
So for example:
```
ytfzt --sub | while read -r line; do
    # parse the line and do whatever you want in here, such as:
    cmd=$( echo "$line" | cut -d':' -f1)
    case $cmd in
        DOWNLOAD) echo "IT WANTS ME TO DOWNLOAD!!!!";;
    esac
done
```

simonhughcom:
The idea is that by using `ytfzf -X`, instead playing the video, downloading the video etc it would just send the command to perform, and the data, to the named pipe, so that the user can decide what he wants to do with it through his own script.

pystardust:
check out `handle_urls`
https://github.com/pystardust/ytfzf/blob/f4a7cb8812798f6af862707812afca2226c0a391/docs/conf.sh#L437
This function decides how the selected urls would be handled.

simonhughcom:
> check out `handle_urls`
> 
> https://github.com/pystardust/ytfzf/blob/f4a7cb8812798f6af862707812afca2226c0a391/docs/conf.sh#L437
> 
> This function decides how the selected urls would be handled.

It doesn't seem to work, I just tried it out and nothing is happening?

Euro20179:
ytfzf will call your implementation or handle_urls, you need to make it do something

Euro20179:
also I had an idea,
we could have a --config option, that way one could have multiple config files which would be very good for a script, especially one that implements things like `handle_urls`

pystardust:
It works something like this

>FILE: conf.sh
```sh
handle_urls () {
	echo "This is the handle function here are the urls -> $*"
}
```

> TERM: after selection
```
~_o $ ytfzf hello
Scraping Youtube...
This is the handle function here are the urls -> https://www.youtube.com/watch?v=YQHsXMglC9A https://www.youtube.com/watch?v=mHONNcZbwDY https://www.youtube.com/watch?v=YxrMNT4SX2E https://www.youtube.com/watch?v=fazMSCZg-mw https://www.youtube.com/watch?v=mv326-zVpAQ
```

Euro20179:
> --config option
nvm this wouldnt work, instead we should have an env variable called `YTFZF_CONFIG_FILE` and it will default to `$YTFZF_CONFIG_DIR/conf.sh` that way anyone could prefix ytfzf with `YTFZF_CONFIG_FILE=` and not have to make a new directory for every config filed

simonhughcom:
Ahh ok, it is working, but not with looping which I had enabled in my config.

Euro20179:
> but not with looping

interesting, this is a but

Euro20179:
actually it seems to work with -l, but since all that's happening in handle_urls is the echoing of the url it echos and immediately goes back into the menu.

Euro20179:
> prefix ytfzf with YTFZF_CONFIG_FILE= and not have to make a new directory for every config filed

apparently I am bad and this is already a thing

simonhughcom:
> actually it seems to work with -l, but since all that's happening in handle_urls is the echoing of the url it echos and immediately goes back into the menu.

So it seems to semi work with -l, but it is very buggy, sometimes it loops, sometimes it does not.
Also the URL it gives is nonsense, the code you gave me prints out `This is the handle function here are the urls -> https://www.youtube.com/watch?v=3 hours ago` the URL is not a valid url. It seems like the video ID is missing.

Euro20179:
what other options do you have set, I can't seem to replicate these problems.

Euro20179:
theoretically if it's printing nonsense urls, it wouldn't be able to play the urls normally because it uses the same variable

```sh
if function_exists "handle_urls"; then
    handle_urls $selected_urls
else
    open_player $selected_urls
fi
```

simonhughcom:
These are my ytfzf configs
https://github.com/simonhughcom/dotfiles/blob/master/ytfzf/.config/ytfzf/conf.sh
I use these and the handle_urls function

pystardust:
> handle_urls function

where is your handle_urls function? The handle_urls has to be mentioned in the conf.sh like so https://github.com/pystardust/ytfzf/blob/f4a7cb8812798f6af862707812afca2226c0a391/docs/conf.sh#L437

simonhughcom:
I have not committed the handle_urls function, I have it on my local copy, its just the one you gave me

Euro20179:
oddly, your config file does break handle_urls and im not sure why, so i'm looking into it

simonhughcom:
Here is my local copy (https://pastebin.com/8VV8Gfpk)

Euro20179:
AHAH, ok i've found it, in `video_info_text` instead of `printf "%-${url_len}.${url_len}s\t" "$shorturl"` change it to `printf "%s" "$shorturl"`, this was changed in an update, I didn't know this was that much of a breaking change, should probably warn people

pystardust:
> was changed in an update

Ah yes, I remember changing this to accommodate bigger videoID's like the playlist's surl.

pystardust:
For legacy support we could add a tabspace at the end, and modify the script where we have used that syntax to extract surl

simonhughcom:
> AHAH, ok i've found it, in `video_info_text` instead of `printf "%-${url_len}.${url_len}s\t" "$shorturl"` change it to `printf "%s" "$shorturl"`, this was changed in an update, I didn't know this was that much of a breaking change, should probably warn people

Yep, that works :D

Euro20179:
I wrote a warning in README.md, I can create a PR, but if we don't remove it now I think we should eventually remove this legacy support.

simonhughcom:
Ok I'm going to close this pull request since everything seems to have been already covered.

PR #293, opened 2021-06-18
feat: silent mode

Euro20179:
if anything else requires the disabling of printing to stdout, we only have to add it to print_info

### DIFF
--- a/docs/conf.sh
+++ b/docs/conf.sh
@@ -11,13 +11,17 @@
 ###################
 
 #this variable should be set somewhere other than here or it will not work
-YTFZF_CONFIG_DIR=$HOME/.config/ytfzf
+
+#YTFZF_CONFIG_DIR=$HOME/.config/ytfzf
+
 
 #the file to source for config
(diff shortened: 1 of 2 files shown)


docs/conf.sh
```diff
@@ -13,6 +13,12 @@
 #this variable should be set somewhere other than here or it will not work
 YTFZF_CONFIG_DIR=$HOME/.config/ytfzf
 
+#the file to source for config
+#the best way to use this would either be to export it in your shell's rc or
+#to run ytfzf as such:
+    #YTFZF_CONFIG_FILE=path/to/config ytfzf ...
+YTFZF_CONFIG_FILE=$YTFZF_CONFIG_DIR/conf.sh
```
pystardust, line 20: This might be confusing, maybe keep it commented and tell them to add it their rc. Having it uncommented  may confuse them if they try editing it

Euro20179: ok, im going to do the same with YTFZF_CONFIG_DIR

PR #294, opened 2021-06-18
docs: document breaking change

Euro20179:


pystardust:
we would need to mention that this is for the `conf.sh`. Also it would go better if the details are in the usage section where we introduce `conf.sh` and add a link to that in the main readme with the heading like `manual intervention for conf.sh`

Euro20179:
how would we implement backwards compatibility for this?

Euro20179:
aight, i've found a solution,
we set `url_len` to 100, then in `preview_img` shorturl is set like this 
```sh
shorturl=${args##*|}
shorturl="$(printf "%s" "$shorturl" | awk '{print $1}')"
```
there are spaces postfixing shorturl, awk will get rid of them  (and for some reason only awk works)

unless there is a problem with this, im going to keep it in this pr, and im going to not put the breaking change on the main page, instead i'll say it's deprecated in USAGE.md

pystardust:
why not remove space using
```
shorturl=${shorturl%% *}
```

Euro20179:
fixed

PR #296, opened 2021-06-19
feat: Add search history and a search history menu

simonhughcom:
This adds a new feature to save the search/query history and a corresponding fzf menu to select old searches.

New flags:
- `ytfzf -q` Open search history menu
- `ytfz -X` Clear search history menu

New Options:
- `enable_search_hist=1` Enable search history saving 
- `enable_search_history_menu=1` Auto use the search history menu

The fzf query will be used with `alt-enter` or if no other selection left, so can be used as an alternative to regular search input.

Also works with looping :D

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -202,9 +202,8 @@ Usage: ytfzf [OPTIONS...] <search-query>;
      -N, --notification                     Send notification when playing video
      -D, --ext-menu                         Use external menu(default dmenu) instead of fzf
      -H, --choose-from-history              Choose from history
-     -x, --clear-history                    Delete history
+     -x, --clear-history=[watch|search]     Delete history
      -q, --search-history                   Open a fzf menu to choose query from search history
-     -X, --clear-search-history             Delete search history
      -m, --audio-only                       Audio only (for music)
      -d, --download                         Download to current directory
      -f                                     Show available formats before proceeding
--- a/ytfzf
+++ b/ytfzf
@@ -1143,13 +1143,29 @@ get_history () {
 }
 
 clear_history () {
-	if [ -e "$history_file" ]; then
-		: > "$history_file"
-		printf "History has been cleared\n"
-	else
-		printf "\033[31mHistory file not found, history not cleared\033[0m\n" >&2
-		exit 1
-	fi
+    case "$1" in
+        search) h_file="${search_history_file}" ;;
+        watch) h_file="${history_file}" ;;
+        *) h_file="${search_history_file}|${history_file}" ;;
+    esac
+
+    exit_code=0
+    IFS="|"
+    for file in $h_file; do
+        if [ "$file" = "$history_file" ]; then
+            name="History"
+        else
+            name="Search History"
+        fi
+        if [ -e "$file" ]; then
+            : > "$file"
+            printf "%s has been cleared\n" "$name"
+        else
+            printf "\033[31m%s file not found, history not cleared\033[0m\n" "$name" >&2
+            exit_code=1
+        fi
+    done
+    exit $exit_code
 }
 
 get_search_history () {
@@ -1165,16 +1181,6 @@ get_search_history () {
 		exit;
 	fi
 	unset hist_data
-}
-
-clear_search_history () {
-	if [ -e "$search_history_file" ]; then
-		: > "$search_history_file"
-		printf "Search history has been cleared\n"
-	else
-		printf "\033[31mSearch history file not found, search history not cleared\033[0m\n" >&2
-		exit 1
-	fi
 }
 
 set_search_history () {
@@ -1392,12 +1398,9 @@ parse_opt () {
 			scrape="history" ;;
 
 		x|clear-history)
-			clear_history && exit ;;
+			clear_history "${optarg}" && exit ;;
         q|search-history)
             search_history_menu ;;
-		X|clear-search-history)
-			clear_search_history && exit ;;
-
 		a|auto-select)
 			auto_select=${optarg:-1}
 			is_non_number "$auto_select" && bad_opt_arg ;;
@@ -1517,7 +1520,7 @@ parse_opt () {
 }
 
 
-while getopts "LhDmdfxXqHaArltSsvNTPYn:U:-:" OPT; do
+while getopts "LhDmdfxqHaArltSsvNTPYn:U:-:" OPT; do
 	# to parse extra-options in conf.sh
 	# when there is no = in OPTARG and it's a longopt, OPTARG will = OPT
 	if [ "$OPT" = "-" ]; then
--- a/ytfzf
+++ b/ytfzf
@@ -1192,11 +1192,11 @@ search_history_menu () {
 
     choice="$( printf "%s\n" "$search_history" | fzf --print-query --no-multi -d '\t' --with-nth=2.. --expect='alt-enter' )"
 
-    query="$( printf "%s" "$choice" | head -n1 )"
+    query="$( printf "%s" "$choice" | sed -n '1p' )"
     [ -z "$query" ] && exit
 
-    key="$( printf "%s" "$choice" | head -n2 )"
-    search="$( printf "%s" "$choice" | head -n3 )"
+    key="$( printf "%s" "$choice" | sed -n '2p' )"
+    search="$( printf "%s" "$choice" | sed -n '3p' )"
 
     if [ -z "$search" ]; then
         search_query="$query"
--- a/ytfzf
+++ b/ytfzf
@@ -1198,10 +1198,13 @@ search_history_menu () {
     fi
 
     query="$( printf "%s" "$choice" | sed -n '1p' )"
-    [ -z "$query" ] && exit
-
     key="$( printf "%s" "$choice" | sed -n '2p' )"
     search="$( printf "%s" "$choice" | sed -n '3p' )"
+
+    # exit if no selection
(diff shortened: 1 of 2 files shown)


ytfzf
```diff
@@ -1143,6 +1152,55 @@ clear_history () {
...
+
+search_history_menu () {
+    get_search_history
+
+    choice="$( printf "%s\n" "$search_history" | fzf --print-query --no-multi -d '\t' --with-nth=2.. --expect='alt-enter' )"
+    query="$( printf "%s" "$choice" | head -n1 )"
+    key="$( printf "%s" "$choice" | head -n2 )"
+    search="$( printf "%s" "$choice" | head -n3 )"
+
+    if [ -z $search ]; then
```
Euro20179, line 1193: quotes are needed around strings in [ ] in case the string is empty.

ytfzf
```diff
@@ -32,6 +32,9 @@ tmp_video_json_file="/tmp/ytfzf-subjson"
 
 #enable/disable history
 enable_hist=${YTFZF_HIST-${enable_hist-1}}
+#enable/disable search history
+enable_search_hist=${YTFZF_SEARCH_HIST-${enable_search_hist-1}}
+enable_search_hist_menu=${YTFZF_SEARCH_HIST_MENU-${enable_search_hist_menu-1}}
```
Euro20179, line 37: this should be 0 by default, when i ran it the first time it told me i had no search history, had to disable it

ytfzf
```diff
@@ -1143,6 +1152,55 @@ clear_history () {
...
+get_search_history () {
+	if [ "$enable_search_hist" -eq 1 ]; then
+		[ -e "$search_history_file" ] || touch "$search_history_file"
+		#gets history data in reverse order (makes it most recent to least recent)
+		hist_data=$( sed '1!G; h; $!d' "$search_history_file" )
+		[ -z "$hist_data" ] && printf "Search history is empty!\n" >&2 && return 1;
+		#removes duplicate values from $history_data
+		search_history=$(printf "%s" "$hist_data" | uniq )
+	else
+		printf "Search history is not enabled. Please enable it to use this option (-H).\n";
```
Euro20179, line 1164: -H should be -q

Euro20179:
i have an idea for clearing history
we have the `-x` option which will clear all history, and the `--clear-history="search|watch"` option which you give either search or watch, and it will clear that history, it uses less options

ytfzf
```diff
@@ -1134,13 +1142,76 @@ get_history () {
...
+    done
+    exit $exit_code
+}
+
+get_search_history () {
+	if [ "$enable_search_hist" -eq 1 ]; then
+		[ -e "$search_history_file" ] || touch "$search_history_file"
+		#gets history data in reverse order (makes it most recent to least recent)
+		hist_data=$( sed '1!G; h; $!d' "$search_history_file" )
+		[ -z "$hist_data" ] && printf "Search history is empty!\n" >&2 && return 1;
```
Euro20179, line 1175: instead of `return 1` this should be `exit 1`.

ytfzf
```diff
@@ -1134,13 +1142,76 @@ get_history () {
...
+		#removes duplicate values from $history_data
+		search_history=$(printf "%s" "$hist_data" | uniq )
 	else
-		printf "\033[31mHistory file not found, history not cleared\033[0m\n" >&2
-		exit 1
+		printf "Search history is not enabled. Please enable it to use this option (-q).\n";
+		exit;
```
Euro20179, line 1180: If you think this should return an exit code of 0, you should also change the exit for when it says History is not enabled.

ytfzf
```diff
@@ -1134,13 +1142,76 @@ get_history () {
...
+
+    if [ "$is_ext_menu" -eq 1 ]; then
+    #when using an external menu, the search history will be done there
+        choice=$( printf "%s\n" "$search_history" | eval "$external_menu" )
+    else
+        choice="$( printf "%s\n" "$search_history" | fzf --print-query --no-multi -d '\t' --with-nth=2.. --expect='alt-enter' --bind='tab:replace-query' )"
+    fi
+
+    query="$( printf "%s" "$choice" | sed -n '1p' )"
+    [ -z "$query" ] && exit
```
Euro20179, line 1201: From testing, I think this line shouldn't exist, when i dont use an ext menu, it just exits here.

ytfzf
simonhughcom: Maybe we should leave it as it is and let the fzf menu open up empty, since it serves a secondary purpose of being able to be used as another form of query input?

ytfzf
simonhughcom: That line is important to make sure that ytfzf exits when nothing has been selected, without it you can have some odd behavior.

Euro20179: hmm, well something needs to be changed i think because it seems to exit even though i'm selecting something.

ytfzf
simonhughcom: Not sure what you mean by
> If you think this should return an exit code of 0

I fixed this exit code.

ytfzf
simonhughcom: Ahh I figured it out, it needs to only exit if both query and search is empty.

Euro20179: instead if checking if query is empty, maybe after query, key and search are defined you can use this line 
```sh
[ -z "$query" ] && [ -z "$key" ] && [ -z "$search" ] && exit
```

also when using an ext_menu, instead of
```sh
choice=$(printf "%s\n" "$search_query" | eval "$external_menu")
```
you could change choice to search_query, and then return on the next line.

ytfzf
Euro20179: well, leaving it as is, is inconsistent with -H when history is empty.

I think it's better to exit, because otherwise you're left with an empty menu which feels buggy, and unintuitive, at least in my opinion. If it said maybe, "Make a search query" it might be ok, but it's still inconsistent with -H, and it would be harder to implement.

Although it is nice to be able to search again from this point, running ytfzf like this would be the same: `ytfzf -q || ytfzf [search]`

simonhughcom: Hmm, history and search history, although similar in many aspect, do have fundamentally different uses.
With history we are only using fzf as a filter for a list of videos, with search history we are using it as a filter for query's as well as an input for new query's to get a new list of videos.
What I trying to say is that there two different aspects of the user pipeline.
Although maybe a good compromise could be a `allow_empty_search_history` option in config?

Euro20179: that would be a good compromise, I see your point of them serving different purposes

my suggestion of how to add it:

```suggestion
if -z "$hist_data" ]; then
    printf "Search history is empty!\n" >&2
    [ $allow_empty_search_history -eq 1 ] && return || exit 1;
fi
```

simonhughcom: I think it might be a better idea to have both histories return 1 and leave it up to the place where the functions are called to handle how to deal with it.

Euro20179: ok

simonhughcom:
I've discovered another bug I have to fix, at the moment if search history menu is enabled no other scraper works. I will fix it later on.

Euro20179:
>  at the moment if search history menu is enabled no other scraper works

so far this is how it's been going for everything, and is what i was talking about when i said it'd be nice to do multiple scrapers at once

to make it more consistent, instead of having `enable_search_hist_menu`, you could have `scrape="search_history"`

ytfzf
```diff
@@ -871,12 +877,15 @@ get_search_query () {
...
 			#when using an external menu, the query will be done there
 			search_query=$( : | eval "$external_menu" )
+        elif [ $enable_search_hist_menu -eq 1 ]; then
+            search_history_menu
 		else
 			#otherwise use the search prompt
 			printf "$search_prompt"
 			read -r search_query
 		fi
 		[ -z "$search_query" ] && clean_up && exit 0
```
Euro20179, line 887: this code block is the else block of the `if [ "$is_ext_menu" -eq 1 ]` check, because when there is no search query it checks if the user is using an external menu. In that case it will prompt the search query there, otherwise it will use the terminal with `read`

Euro20179: also you might want to move `set_search_history` outside of the if check, because if there is no search and the user just runs, `ytfzf` then gets the search prompt, it won't store the search.

simonhughcom: > also you might want to move `set_search_history` outside of the if check, because if there is no search and the user just runs, `ytfzf` then gets the search prompt, it won't store the search.

`set_search_history` is outside of the if check?

Euro20179: cool, if you're ready for it to be merged add a reaction to this, cause it looks good to me

simonhughcom: I think I'll add a few more basic user options before merge.

simonhughcom: Do you want me to add anything to the docs? Or add the options to the example conf.sh?
Otherwise I am ready for the PR to be merged.

Euro20179: you should add 
`enable_search_hist`, `enable_search_hist_menu`, and `allow_empty_search_hist` to the `OPT VARIABLES` section in `docs/conf.sh`
and add
`search_history_prompt`, and `search_history_file` to `MSIC/Other` or near where `useragent` is.

Euro20179: > set_search_history is outside of the if check?

`set_search_history` needs to be called the line before `function_exists "on_get_search"...` because it doesn't save when you search like `ytfzf search`

PR #298, opened 2021-06-21
fix: Add missing is_download config option

simonhughcom:
Not sure why this was missing, but this adds the `is_download` config option.


Euro20179:
can you move the conf.sh part to the OPT VARIABLES section

PR #299, opened 2021-06-22
feat: pt thumbnails; fix: on_video_data_gotten not called in scrape_pt

Euro20179:


ytfzf
```diff
@@ -975,16 +980,13 @@ format_user_selection () {
...
+		case ${#surl} in
 					# youtube video
-					11) selected_urls=$selected_urls$new_line'https://www.youtube.com/watch?v='$surl ;;
-					34) selected_urls=$selected_urls$new_line'https://www.youtube.com/playlist?list='$surl ;;
-					*) continue ;;
-				esac
-				;;
+			11) selected_urls=$selected_urls$new_line'https://www.youtube.com/watch?v='$surl ;;
+			34) selected_urls=$selected_urls$new_line'https://www.youtube.com/playlist?list='$surl ;;
+			36)
```
pystardust, line 987: oh so for peertube the pre-url part  is not consistent?

pystardust (approved):

ytfzf
Euro20179: in order to do thumbnails, i had to store the video id for the file path to download the thumbnail to. I am also using the video id here which will always be a UUID of 36 characters.

PR #305, opened 2021-06-28
Url watch history

simonhughcom:
This pr adds a scrape function for urls to get their video data and save it to watch history.

I am using `youtube-dl -J` to get the video json, however the data it returns is not in the same format as the other scrapers (e.g. duration in seconds instead of HH:MM:SS), this however does not seem to affect the functionality.

NOTE: everything works when `sort_videos_data` is disabled, however there seems to be a bug in the `sort_video_data_fn` function where the first line does not get returned.
This bug effects the other functions of ytfzf as well, such as the first result of `scrape_yt` gets removed.

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -915,7 +915,7 @@ scrape_url () {
 	    exit 1
 	fi
 
-	function_exists "on_video_data_gotten" && on_video_data_gotten "$videos_data" "$videos_json" "$yt_json"
+	function_exists "on_video_data_gotten" && on_video_data_gotten "$videos_data" "$videos_json" "$url_json"
 
 	[ "$sort_videos_data" -eq 1 ] && videos_data=$(printf "%s" "$videos_data" | sort_video_data_fn)
 
--- a/ytfzf
+++ b/ytfzf
@@ -892,8 +892,6 @@ scrape_url () {
 		print_error "ERROR[#02]: Couldn't find data on site.\n"
 		exit 1
 	fi
-    # printf "%s" "$url_json" | jq
-    # exit
 
 	#gets a list of videos
     videos_json=$(printf "%s" "$url_json" | jq '[{
--- a/ytfzf
+++ b/ytfzf
@@ -919,10 +919,14 @@ scrape_url () {
 
 	[ "$sort_videos_data" -eq 1 ] && videos_data=$(printf "%s\n" "$videos_data" | sort_video_data_fn)
 
+    unset url_json
+}
+
+scrape_url_with_thumb () {
+    scrape_url $*
 	[ $show_thumbnails -eq 1 ] && download_thumbnails "$videos_json"
 	# wait for thumbnails to download
 	wait
-    unset url_json
 }
 
 ############################
--- a/ytfzf
+++ b/ytfzf
@@ -917,7 +917,7 @@ scrape_url () {
 
 	function_exists "on_video_data_gotten" && on_video_data_gotten "$videos_data" "$videos_json" "$url_json"
 
-	[ "$sort_videos_data" -eq 1 ] && videos_data=$(printf "%s" "$videos_data" | sort_video_data_fn)
+	[ "$sort_videos_data" -eq 1 ] && videos_data=$(printf "%s\n" "$videos_data" | sort_video_data_fn)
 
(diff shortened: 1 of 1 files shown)


ytfzf
```diff
@@ -878,6 +878,52 @@ scrape_yt () {
...
+
+	#checks if it's empty in case it was defined in a config function eg: on_get_search
+	[ -z "$videos_data" ] && videos_data=$(get_video_data "$videos_json")
+	#if there aren't videos
+	if [ -z "$videos_data" ]; then
+	    printf "No results found. Try different keywords.\n" >&2
+	    exit 1
+	fi
+
+	function_exists "on_video_data_gotten" && on_video_data_gotten "$videos_data" "$videos_json" "$yt_json"
```
Euro20179, line 918: instead of `$yt_json` it should be `$url_json`

ytfzf
```diff
@@ -878,6 +878,52 @@ scrape_yt () {
...
+
+	[ "$sort_videos_data" -eq 1 ] && videos_data=$(printf "%s" "$videos_data" | sort_video_data_fn)
+
+	[ $show_thumbnails -eq 1 ] && download_thumbnails "$videos_json"
+	# wait for thumbnails to download
+	wait
```
Euro20179, line 924: lines 922-924 are unnecessary because these thumbnails wont be seen anyway.

simonhughcom: The thumbnails can get used in the history menu, by simply using `scrape_url` in `get_history`, but it does make using urls in cmd line less efficient. I'll separate these into two functions to allow the best of both.

simonhughcom:
There is another minor problem, currently the description is too long and being covered by the thumbnail in the history menu.

Euro20179:
A couple things i noticed:
1. since there is no `videos_json` with history, the thumbnail menu *just* has a thumbnails.
2. playlists take FOREVER to download in history.

But, since history without -t works the same im not sure how big of an issue this is, since we could maybe add this line in `get_history` when thumbnails are enabled.
`printf "\033[33mWarning: you have thumbnails enabled with history, if this takes too long try using -H without thumbnails\033[0m >&2`

simonhughcom:
1. I added a videos_json for get_history with the last 2 commits. I'm not sure what you mean.
2. Its `youtube-dl` downloading the json that takes so long with playlists since ts downloading the data and thumbnails for all videos, however we only really need some of the data and one thumbnail. I want to see if I can figure out a way to only get the first thumbnail and use that for the playlist.

Euro20179:
> I'm not sure what you mean.

with -t the title, description, length, etc.. dont show on the side, I'll check again in a bit to make sure it's not there because I might just be crazy haha.

simonhughcom:
They are not showing currently for playlists, they are for normal videos.

simonhughcom:
So I have come up with 4 solutions to the problem, each of them have advantages and disadvantages:

1. we exclude playlist urls from being saved to history, this is the easiest solution but it does mean no playlist in history.

2. we single out playlists from history and use a alternative method for retrieving the playlist data, more complex and a little hacky, this works reasonably well when there are lots of videos and a few playlists in history, but not so well when there are lots of playlists.

3. we single out playlists again but don't retrieve the data and just have a placeholder image.

4. Under my view the best solution, but it also requires some work, we save the thumbnail URL and description in history and use that.

Euro20179:
> we single out playlists again but don't retrieve the data and just have a placeholder image.

There is technically a default image already called `default_thumb.png` in `~/.config/ytfzf` or wherever the config directory is. So this would be a decent solution. IMO this is the best solution.

Changing the layout of the history file would probably break external scripts, and we should do it cautiously.

simonhughcom:
Tbh even with excluding playlists, if we get to more then 10 videos in history, it gets unbearably slow, I thinks its because youtube-dl is making a separate request for every video.

I don't think that there is an efficient way of doing it apart from doing it locally.

Euro20179:
maybe we scrap getting thumbnails since it's so slow

simonhughcom:
Completely or just from get_history?

Euro20179:
just from get_history

simonhughcom:
Yes that is probably a good idea, I'm not sure if there is a plan for ytfzf v2, but if we ever do that we can revisit this.

Euro20179:
I dont think there are plans for v2. But we could revisit this

PR #310, opened 2021-06-30
Fix: remove duplicate search history based on title

simonhughcom:
Ignores the date and time and only uses title to compare search history


Euro20179:
This doesn't seem to do anything?

simonhughcom:
It should remove duplicate search history.
Make sure that you search for the same thing twice directly after each other, and when using `ytfzf -q` it should only return one instance of the search.

Euro20179:
ah, ok

PR #311, opened 2021-06-30
Add a persistent mode

simonhughcom:
This PR ads a new option `enable_persistent=1` which launches `ytfzf` in persistent mode, in this mode the fzf --bind option is used to execute and open the player without exiting fzf.

This has the advantage of:
- no looping, fzf will just persist
- unlike looping, the pointer will stay at the last position (far less annoying)

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -96,6 +96,9 @@ key_binds="${key_binds-alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter,enter
 # these keys allow the use of key_binds without leaving fzf
 persistent_key_binds="${persistent_key_binds-enter,alt-enter,alt-v,alt-d,alt-o,double-click}"
 
+# Depreciation warning for shortcuts
+[ -n "${shortcuts+shortcuts_is_defined}" ] && printf "\033[33mWarning: shortcuts is defined, use the 'key_binds' variable instead\033[0m\n" >&2
+
 search_prompt=${search_prompt-Search: }
 
 #used when getting the html from youtube
--- a/ytfzf
+++ b/ytfzf
@@ -954,27 +954,30 @@ get_search_query () {
     set_search_history
 	function_exists "on_get_search" && on_get_search "$search_query"
 }
+
+persistent_shortcuts () {
+    binds=
+    # make sure every key code is surrounded by commas
+    shortcuts=",$shortcuts,"
+    IFS=','
+    for key in $persistent_shortcuts; do
+        bind_command=":execute-silent(sh $0 -U \"fzf_bind_command $key \"{+f})"
+        binds="${binds},${key}${bind_command}"
+        # Remove used keys from shortcut
+        shortcuts="$( printf "%s" "$shortcuts" | sed "s/,$key,/,/g" )"
+    done
+    unset IFS
+    shortcuts="${shortcuts%,*}"
+    shortcuts="${shortcuts#*,}"
+}
+
 #> To select videos from videos_data
 user_selection () {
+    # format persistent shortcuts for fzf
+    persistent_shortcuts
+
 	#remove subscription separators
 	videos_data_clean=$(printf "%s" "$videos_data" | sed "/.*$tab_space$/d")
-    
-    ## PERSISTENT SHORTCUTS
-    if [ "$enable_loop" -eq 1 ]; then
-        loop_binds=
-        # make sure every key code is surrounded by commas
-        shortcuts=",$shortcuts,"
-        IFS=','
-        for key in $persistent_shortcuts; do
-            bind_command=":execute-silent(sh $0 -U \"fzf_bind_command $key \"{+f})"
-            loop_binds="${loop_binds},${key}${bind_command}"
-            # Remove used keys from shortcut
-            shortcuts="$( printf "%s" "$shortcuts" | sed "s/,$key,/,/g" )"
-        done
-        unset IFS
-        shortcuts="${shortcuts%,*}"
-        shortcuts="${shortcuts#*,}"
-    fi
 
 	#$selected_data is the video the user picked
 	#picks the first n videos
@@ -994,7 +997,7 @@ user_selection () {
 		[ "$thumb_disp_method" = "ueberzug" ] && start_ueberzug
 		#thumbnails only work in fzf, use fzf
         menu_command="fzf -m --tabstop=1 \
-        --bind 'change:top$loop_binds' --delimiter=\"$tab_space\" \
+        --bind 'change:top$binds' --delimiter=\"$tab_space\" \
 		--nth=1,2 --expect='$shortcuts' $FZF_DEFAULT_OPTS \
         --layout=reverse --preview \"sh $0 -U 'preview_img '{}\" \
         	--preview-window \"$PREVIEW_SIDE:50%:noborder:wrap\""
--- a/ytfzf
+++ b/ytfzf
@@ -472,12 +472,13 @@ start_ueberzug () {
     [ -e $FIFO ] || { mkfifo "$FIFO" || exit 1 ; }
     ueberzug layer --parser json --silent < "$FIFO" &
     exec 3>"$FIFO"
+    ueberzug_pid="$!"
 }
 stop_ueberzug () {
     exec 3>&-
     rm "$FIFO" > /dev/null 2>&1
     # make sure ueberzug is really dead
-    ps aux | grep 'ueberzug' | awk '{print $2}' | sed '1q' | xargs kill
+    kill "$ueberzug_pid"
 }
 
 preview_img () {
--- a/ytfzf
+++ b/ytfzf
@@ -969,7 +969,10 @@ persistent_shortcuts () {
     # Build fzf --bind command
     IFS=','
     for key in $persistent_shortcuts; do
-        bind_command=":execute-silent(sh $0 -U \"fzf_bind_command $key \"{+f})"
+        # not sure why, but this won't work with `execute-silent`
+        # sleep is necessary to give the command enough time to execute before
+        # the fzf tmp file gets removed
+        bind_command=":execute(sh -c \"$0 -U \\\"fzf_bind_command $key {+f}\\\"\" >/dev/null 2>&1 & sleep 0.1)"
         binds="${binds},${key}${bind_command}"
         # Remove used keys from shortcut
         shortcuts="$( printf "%s" "$shortcuts" | sed "s/,$key,/,/g" )"
@@ -1018,7 +1021,7 @@ user_selection () {
 		[ "$thumb_disp_method" = "ueberzug" ] && start_ueberzug
 		#thumbnails only work in fzf, use fzf
         menu_command="fzf -m --tabstop=1 \
-        --bind 'change:top$ignore_binds$binds' --delimiter=\"$tab_space\" \
+        --bind 'change:top"$ignore_binds$binds"' --delimiter=\"$tab_space\" \
 		--nth=1,2 --expect='$shortcuts' $FZF_DEFAULT_OPTS \
         --layout=reverse --preview \"sh $0 -U 'preview_img '{}\" \
         	--preview-window \"$PREVIEW_SIDE:50%:noborder:wrap\""
--- a/ytfzf
+++ b/ytfzf
@@ -969,7 +969,7 @@ user_selection () {
 		#thumbnails only work in fzf, use fzf
 		menu_command="fzf -m --tabstop=1 --bind change:top --delimiter=\"$tab_space\" \
 		--nth=1,2 --expect='$shortcuts' $FZF_DEFAULT_OPTS \
-		--layout=reverse --preview \"sh $0 -U {}\" \
+        --layout=reverse --preview \"sh $0 -U 'preview_img '{}\" \
         	--preview-window \"$PREVIEW_SIDE:50%:noborder:wrap\""
 		selected_data=$( title_len=200 video_menu "$videos_data" )
 		[ "$thumb_disp_method" = "ueberzug" ] && stop_ueberzug
@@ -1430,6 +1430,14 @@ bad_opt_arg () {
 	exit 3
 }
 
+parse_internal_opt (){
+    arg=${*%% *}
+    val=${*#* }
+    case "$arg" in
+        preview_img*) preview_img "$val"; exit;;
+    esac
+}
+
 parse_opt () {
 	#the first arg is the option
 	opt=$1
@@ -1551,8 +1559,8 @@ parse_opt () {
 			link_count="$optarg"
 			is_non_number "$link_count" && bad_opt_arg "$link_count" "-n" ;;
 
-		U) 	preview_img "$optarg"; exit;
-			# This option is reserved for the script, to show image previews
+		U) 	parse_internal_opt "$optarg";
+			# This option is reserved for internal scripts
 			# Not to be used explicitly
 			;;
 
--- a/ytfzf
+++ b/ytfzf
@@ -1512,7 +1512,7 @@ parse_internal_opt (){
     val=${*#* }
     case "$arg" in
         preview_img*) preview_img "$val"; exit;;
-        fzf_bind_command*) fzf_bind_command "$val";
+        fzf_bind_command*) fzf_bind_command "$val";;
     esac
 }
 
--- a/ytfzf
+++ b/ytfzf
@@ -1038,7 +1038,7 @@ fzf_bind_command () {
     args=$*
     selected_data="$( cat "${args#* }" )"
     format_user_selection
-    selected_key=${*%% *}
+    selected_key=${args%% *}
 	handle_key_binds
 	case "$?" in
 	    1)
@@ -1504,8 +1504,9 @@ bad_opt_arg () {
 }
 
 parse_internal_opt (){
-    arg=${*%% *}
-    val=${*#* }
+    arg=$*
+    val=${arg#* }
+    arg=${arg%% *}
     case "$arg" in
         preview_img*) preview_img "$val"; exit;;
         fzf_bind_command*) fzf_bind_command "$val";;
--- a/ytfzf
+++ b/ytfzf
@@ -89,14 +89,12 @@ pid_file="$cache_dir/.pid"
 [ -d "$thumb_dir" ] || mkdir -p "$thumb_dir"
 
 #> config settings
-#list of shortcuts to use in fzf
-#handle_custom_shortcuts will be called if defined (handle_custom_shortcuts can call handle_shortcuts to handle normal shortcuts)
-# These binds are used for standard select function
-select_binds="${select_binds-enter,double-click}"
-# shortcuts are used for alternative functions apart from select
-shortcuts="${shortcuts-alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter}"
-# these keys allow the use of shortcuts or select binds without leaving fzf
-persistent_shortcuts="${persistent_shortcuts-enter,alt-enter,alt-v,alt-d,alt-o,double-click}"
+#handle_custom_key_binds will be called if defined (handle_custom_key_binds can call handle_key_binds to handle normal key binds)
+# key_binds sets the bindings to use for fzf
+# any undefined binds will be used to select
+key_binds="${key_binds-alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter,enter,double-click}"
+# these keys allow the use of key_binds without leaving fzf
+persistent_key_binds="${persistent_key_binds-enter,alt-enter,alt-v,alt-d,alt-o,double-click}"
 
 search_prompt=${search_prompt-Search: }
 
@@ -409,7 +407,7 @@ format_menu () {
 	if [ "$is_ext_menu" -eq 0 ]; then
 		#dep_ck fzf here because it is only necessary to use here
 		dep_ck "fzf"
-		menu_command='column -t -s "$tab_space" | fzf -m --bind change:top --tabstop=1 --layout=reverse --delimiter="$tab_space" --nth=1,2 --expect="$shortcuts" $FZF_DEFAULT_OPTS'
+		menu_command='column -t -s "$tab_space" | fzf -m --bind change:top --tabstop=1 --layout=reverse --delimiter="$tab_space" --nth=1,2 --expect="$key_binds" $FZF_DEFAULT_OPTS'
 		format_fzf
 	else
 		# Dmenu doesn't render tabs so removing it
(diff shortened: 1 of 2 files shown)


Euro20179:
what is the benefit of not looping?

Having the pointer stay in the last position is good, but i feel that there is a better way to do this, and instead of having an option just do it by default with -l

simonhughcom:
Technically is more efficient, but the pointer is the main reason.

So you want to have looping do the persistent fzf stuff? Its fine by me.

Euro20179:
yes technically the benefit of not looping is more efficient.

I am currently trying to write a way to save the cursor position after selecting a video

simonhughcom:
I just noticed that multi select does not work yet, I'll fix it later.

Euro20179:
another problem is that `enter` isn't the only way to select a video, `double-click` is another default, so in my version i had this variable `select_video_in_fzf="${select_video_in_fzf-enter,double-click}"` which can be modified in case they select videos in a  different way.

ytfzf
```diff
@@ -464,10 +468,13 @@ start_ueberzug () {
     [ -e $FIFO ] || { mkfifo "$FIFO" || exit 1 ; }
     ueberzug layer --parser json --silent < "$FIFO" &
     exec 3>"$FIFO"
+    ueberzug_pid="$!"
```
Euro20179, line 471: $! doesn't seem to work in dash.

ytfzf
```diff
@@ -464,10 +468,13 @@ start_ueberzug () {
...
 }
 stop_ueberzug () {
     exec 3>&-
     rm "$FIFO" > /dev/null 2>&1
+    # make sure ueberzug is really dead
+    /bin/kill $ueberzug_pid
```
Euro20179, line 477: if there are 2 instances of ytfzf running, this will break thumbnails on all other instances

Euro20179, line 477: instead of this maybe use this line or a variation:
```sh
ps aux | grep ueberzug | awk '{print $2}' | sed 1q | xargs kill
```

simonhughcom:
> another problem is that `enter` isn't the only way to select a video, `double-click` is another default, so in my version i had this variable `select_video_in_fzf="${select_video_in_fzf-enter,double-click}"` which can be modified in case they select videos in a different way.

I can add that variable no problem.
Just curious, how does your version save the cursor position?

Euro20179:
It's easier to show the diff
```diff
--- a/ytfzf
+++ b/ytfzf
@@ -91,6 +91,7 @@ pid_file="$cache_dir/.pid"
 #> config settings
 #list of shortcuts to use in fzf
 #handle_custom_shortcuts will be called if defined (handle_custom_shortcuts can call handle_shortcuts to handle normal shortcuts)
+select_video_in_fzf="enter,double-click"
 shortcuts="${shortcuts-alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter}"
 search_prompt=${search_prompt-Search: }
 #used when getting the html from youtube
@@ -399,7 +400,7 @@ format_menu () {
        if [ "$is_ext_menu" -eq 0 ]; then
                #dep_ck fzf here because it is only necessary to use here
                dep_ck "fzf"
-               menu_command='column -t -s "$tab_space" | fzf -m --bind change:top --tabstop=1 --layout=reverse --delimiter="$tab_space" --nth=1,2 --expect="$shortcuts" $FZF_DEFAULT_OPTS'
+               menu_command='column -t -s "$tab_space" | fzf -m --bind change:top --tabstop=1 --layout=reverse --delimiter="$tab_space" --nth=1,2 --expect="$shortcuts,$select_video_in_fzf" $FZF_DEFAULT_OPTS'
                format_fzf
        else
                # Dmenu doesn't render tabs so removing it
@@ -924,7 +925,7 @@ user_selection () {
                [ "$thumb_disp_method" = "ueberzug" ] && start_ueberzug
                #thumbnails only work in fzf, use fzf
                menu_command="fzf -m --tabstop=1 --bind change:top --delimiter=\"$tab_space\" \
-               --nth=1,2 --expect='$shortcuts' $FZF_DEFAULT_OPTS \
+               --nth=1,2 --expect='$shortcuts,$select_video_in_fzf' $FZF_DEFAULT_OPTS \
                --layout=reverse --preview \"sh $0 -U {}\" \
                --preview-window \"$PREVIEW_SIDE:50%:noborder:wrap\""
                selected_data=$( title_len=200 video_menu "$videos_data" )
@@ -943,11 +944,13 @@ handle_shortcuts () {
     #creates splits the variable shortcuts by , and assigns a variable to each string
     #ei: $urls_shortcut will be equal to the first keybind
     IFS="," read -r \
-       urls_shortcut title_shortcut open_browser_shortcut watch_shortcut \
+       _ urls_shortcut title_shortcut open_browser_shortcut watch_shortcut \
        download_shortcut audio_shortcut search_shortcut detach_shortcut _ <<-EOF
        $shortcuts
        EOF
 
+    last_mouse_pos="$(printf "%s" "$videos_data" | grep -n "$(printf "%s\n" "$selected_data" | awk -F "\t|" '{print $NF}')" | cut -d ':' -f1 | sed 1q)"
+
     case $selected_key in
        "$urls_shortcut") printf "%s\n" $selected_urls; return 1 ;;
        "$title_shortcut") 
@@ -1665,4 +1668,5 @@ while true; do
                scrape_fn
                continue
        fi
```

But the core of it is, when you select a video, it selects the line that the video is on, then when fzf opens again it *would* automatically put you back on that line, but it seems fzf doesn't support that, i was thinking about doing some trickery with `printf "\033[B"` (down arrow key press i think) but that wouldn't work either and it's probably terminal specific.

simonhughcom:
Yea I was thinking of a similar approach, but after looking at the fzf readme they where doing something like this using the bind command, which seemed to be a better approach.

Euro20179:
ah ok

ytfzf
```diff
@@ -150,6 +150,10 @@ trending_tab="${trending_tab-}"
 sort_name="${sort_name-}"
 #to detach the video player from the terminal
 detach_player=${detach_player-0}
+
+# in loop mode the player always needs to detach
+[ "$enable_loop" -eq 1 ] && detach_player=1
```
Euro20179, line 155: add the `detach_player=1` part in the `parse_opt` function where `-l` is as well as here.

ytfzf
```diff
@@ -949,6 +955,8 @@ get_search_query () {
 user_selection () {
 	#remove subscription separators
 	videos_data_clean=$(printf "%s" "$videos_data" | sed "/.*$tab_space$/d")
+    
+    [ "$enable_loop" -eq 1 ] && loop_binds=",enter:execute-silent(sh $0 -U \"selected_data_from_file \"{+f})"
```
Euro20179, line 959: add some variable maybe `fzf_select_bindings="${fzf_select_bindings-enter,double-click}"` instead of hard coding enter here.

ytfzf
```diff
@@ -982,6 +990,17 @@ user_selection () {
...
+    selected_data="$( cat "$*" )"
+    [ -z "$selected_data" ] && exit
+	format_user_selection
+	get_video_format
+	get_sub_lang
+	play_url
+	save_before_exit
+    exit
+}
+
```
Euro20179, line 1003: instead of this maybe we just make this function set `selected_data` to the data in the file, then continue the main loop like normal (but skipping the `user_selection` bit), i'm not exactly sure what the best way to check to see if `user_selection` should be skipped though.

IMO, this function just feels very repetitive and inefficient, that's what I'm trying to solve.

simonhughcom:
I'm working on the `fzf_select_bindings`, i'm trying to split it into 2 variables `fzf_select_bindings_play` and `fzf_select_bindings_download` so a user can have separate bindings for playing and downloading.

The reason for it being in a file is because of the way fzf handles the output for multi selections, without a file it just space separates them making it difficult to parse. In a file it line separates them. Its the best solution I could come up with.

We could put the main loop into its own function and call it in `selected_data_from_file` and after the user selection.
The other problem however is that we need to not have a loop when `selected_data_from_file` is called.

Euro20179:
> The reason for it being in a file is because of the way fzf handles the output for multi selections, without a file it just space separates them making it difficult to parse. In a file it line separates them. Its the best solution I could come up with.

Ah, ok.

> I'm working on the fzf_select_bindings, i'm trying to split it into 2 variables fzf_select_bindings_play and fzf_select_bindings_download so a user can have separate bindings for playing and downloading.

I don't see a need to do this, because this variable is mainly just so that when you select something it triggers the functions. But maybe there is a need im not sure.

> and after the user selection

Why does it need to be called after user selection?

simonhughcom:
> I don't see a need to do this, because this variable is mainly just so that when you select something it triggers the functions. But maybe there is a need im not sure.

Its so the use can have two separate key bindings, one for download and one for playing. Atm this is accomplished with shortcuts, but the problem with these is that they require exiting fzf.

> Why does it need to be called after user selection?

For when the user launches `ytfzf`, otherwise nothing happens.

Euro20179:
> With after user selection I mean so that the main loop gets called like normal.

Ohh, ok.

> Its so the use can have two separate key bindings, one for download and one for playing. Atm this is accomplished with shortcuts, but the problem with these is that they require exiting fzf.

ah i see, I think it might be better to make shortcuts accomplish this by setting `enable_loop=1` in the shortcuts.

> With after user selection I mean so that the main loop gets called like normal.

Ok.

simonhughcom:
> ah i see, I think it might be better to make shortcuts accomplish this by setting `enable_loop=1` in the shortcuts.
The problem with shortcuts is that they use the `--expect` function of fzf, which exits fzf. So we would loose the pointer position etc

I'm trying to do this with `--bind` so that we stay in fzf and keep pointer position for playing vids, downloading (maybe even some other stuff).

Euro20179:
> which exits fzf

Ok.

simonhughcom:
I hope you don't mind, but I changed the approach to the problem to what I think is a much better way of using binds.

Instead of having a separate way of dealing with binds, the `persistent_shortcuts` variable simple allows you to define keys that will use shortcuts without leaving fzf via `fzf --bind`, that way we can reuse `handle_shortcuts`.

This also works without looping, so I thinks we ought to remove that restriction, if a person wishes to not use this, he can simple do `persistent_shortcuts=""`

ytfzf
```diff
@@ -956,7 +959,22 @@ user_selection () {
...
+        for key in $persistent_shortcuts; do
+            bind_command=":execute-silent(sh $0 -U \"fzf_bind_command $key \"{+f})"
+            loop_binds="${loop_binds},${key}${bind_command}"
+            # Remove used keys from shortcut
+            shortcuts="$( printf "%s" "$shortcuts" | sed "s/,$key,/,/g" )"
+        done
+        unset IFS
+        shortcuts="${shortcuts%,*}"
+        shortcuts="${shortcuts#*,}"
+    fi
```
Euro20179, line 977: This should probably be it's own function, maybe `generate_loop_binds`

ytfzf
```diff
@@ -990,10 +1009,19 @@ user_selection () {
 	unset videos_data_clean
 }
 
-selected_data_from_file () {
-    selected_data="$( cat "$*" )"
-    [ -z "$selected_data" ] && exit
-	format_user_selection
+fzf_bind_command () {
+    file=${*#* }
```
Euro20179, line 1013: according to `shellcheck` `${*}` isn't posix, this is how I would fix it
```sh
args=$*
file=${args#* }
```

ytfzf
```diff
@@ -92,9 +92,12 @@ pid_file="$cache_dir/.pid"
 #list of shortcuts to use in fzf
 #handle_custom_shortcuts will be called if defined (handle_custom_shortcuts can call handle_shortcuts to handle normal shortcuts)
 shortcuts="${shortcuts-alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter}"
+# these keys allow the use of shortcuts without leaving fzf
+persistent_shortcuts="${fzf_select_bindings_play-enter,alt-enter,alt-v,alt-d,alt-o}"
```
Euro20179, line 96: maybe rename `fzf_select_bindings_play` to `fzf_bind_shortcuts`

`fzf_select_bindings_play` is a bit long and has a bit of an odd name.

simonhughcom:
Do you want me to remove the loop restriction and just make it a default feature as part of shortcuts?

I would also have to rewrite `handle_shortcuts` to accept `enter` and `double-click` etc. I think we should rename `handle_shortcuts` to `handle_binds` and have every user action go through there.

Euro20179:
> just make it a default feature as part of shortcuts?

this is probably a good idea.

Euro20179:
key inputs seem to not work after selecting a video

simonhughcom:
> key inputs seem to not work after selecting a video
> 
> (in case you cannot replicate this -s is enabled)

Are you talking about fzf freezing when you play a vid?
That's because fzf is waiting for the command to finish, if you quit the video fzf will work again, whatever process fzf is running through bind needs to be detached to not hold fzf up.

Euro20179:
> That's because fzf is waiting for the command to finish

This should probably be fixed, unless its unfixable.

simonhughcom:
This latest commit seems to solve the problem by sending the task to background, however I had to use `sleep` to give enough time to read the tmp file before fzf removes it, `sleep 0.1` seems to be enough on my system, but I am not sure about other systems.

simonhughcom:
Atm we have the `select_binds` conf var, however it is superfluous. The only reason why it is there is to have shortcuts be backwards compatible, but it could and possibly should really all be one config option, but it would break backwards compatibility with shortcuts in a minor way.

I think we should rename `shortcuts` to `binds` and have it handle all binds, it would be similar to `shortcuts` (i.e. the first x keys are already assigned, and you can assign more in conf) but any unassigned keys would be used to accept.

what do you think?

simonhughcom:
Oh and I need to write the bind commands for fzf without thumbs. I'll do that later.

Euro20179:
> but any unassigned keys would be used to accept.

idk about this.

It is also good to keep things backwards compatible, but eventually things should change, since `shortcuts=` is new i think renaming it to `binds=` as long as this update merges to master soon-ish shouldn't be a large issue, and as long as we say.

We could check if `select_binds` is defined then throw an error 
```sh
if type select_binds; then
    print_error "select_binds has been removed, [say the new way]"
    exit 3
fi
```

simonhughcom:
> We could have all unused binds used for select, then select_binds could be used to define a specific set of bindings

That's how it works right now, but that is my point, we don't really need it.

if you have a look at line 1062
``` shell
handle_shortcuts () {
    [ -z "$selected_key" ] && return 0

    printf ",%s," "$select_binds" | grep -q ",$selected_key," && return 0

    #creates splits the variable shortcuts by , and assigns a variable to each string
    #ei: $urls_shortcut will be equal to the first keybind
```

currently we are just grepping for `select_binds` for accept, however like I said, I think this is superfluous since any key that will not get grepped here or trigger a case statement in the `handle_shortcuts` function gets used for accept anyway.

Therefore if we remove `select_binds` and only have `shortcuts` (which I think should then be renamed to `binds`) we would have to warn people to append their preferred accept keys to the end of `shortcuts`/`binds`

Euro20179:
ok, that sounds good

ytfzf
```diff
@@ -967,51 +1018,74 @@ user_selection () {
...
 	fi
 	unset videos_data_clean
 }
 
-handle_shortcuts () {
+fzf_bind_command () {
+    args=$*
+    selected_data="$( cat "${args#* }" )"
+    format_user_selection
+    selected_key=${*%% *}
```
pystardust, line 1042: @Euro20179 this isn't POSIX complaint right? store $* in a variable and then trim the variable

ytfzf
```diff
@@ -89,12 +89,16 @@ pid_file="$cache_dir/.pid"
 [ -d "$thumb_dir" ] || mkdir -p "$thumb_dir"
 
 #> config settings
-#list of shortcuts to use in fzf
-#handle_custom_shortcuts will be called if defined (handle_custom_shortcuts can call handle_shortcuts to handle normal shortcuts)
-shortcuts="${shortcuts-alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter}"
+#handle_custom_key_binds will be called if defined (handle_custom_key_binds can call handle_key_binds to handle normal key binds)
+# key_binds sets the bindings to use for fzf
+# any undefined binds will be used to select
+key_binds="${key_binds-alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter,enter,double-click}"
```
pystardust, line 95: this would cause backward compatibility problems as `$shortcuts` are set in conf.sh. So the new `key_binds` should use that instead

pystardust (approved):

ytfzf
```diff
@@ -468,6 +477,8 @@ start_ueberzug () {
 stop_ueberzug () {
     exec 3>&-
     rm "$FIFO" > /dev/null 2>&1
+    # make sure ueberzug is really dead
+    ps aux | grep 'ueberzug' | awk '{print $2}' | sed '1q' | xargs kill
```
pystardust, line 481: the user might have other programs using ueberzug

ytfzf
Euro20179: yes store $* in a variable then use the variable

ytfzf
Euro20179: yes, we were going to add a warning to change it, `type shortcuts && printf "Error...."`

ytfzf
Euro20179: true, not sure how to fix that.

ytfzf
```diff
@@ -399,7 +407,8 @@ format_menu () {
 	if [ "$is_ext_menu" -eq 0 ]; then
 		#dep_ck fzf here because it is only necessary to use here
 		dep_ck "fzf"
-		menu_command='column -t -s "$tab_space" | fzf -m --bind change:top --tabstop=1 --layout=reverse --delimiter="$tab_space" --nth=1,2 --expect="$shortcuts" $FZF_DEFAULT_OPTS'
+		menu_command="column -t -s '$tab_space' | fzf -m --bind 'change:top"$ignore_binds$binds"' --tabstop=1 --layout=reverse --delimiter='$tab_space' --nth=1,2 --expect='$key_binds' $FZF_DEFAULT_OPTS"
+        echo "$menu_command"
```
Euro20179, line 411: leftover debugging?

ytfzf
```diff
@@ -1430,6 +1504,15 @@ bad_opt_arg () {
 	exit 3
 }
 
+parse_internal_opt (){
+    arg=${*%% *}
+    val=${*#* }
```
Euro20179, line 1509: set $* to a variable first then use it in brace expansion

ytfzf
simonhughcom: I think I should go back to using `$!`.
@Euro20179 seems to think it does not work in dash, but it works with my dash, so it must be a problem on your end.

ytfzf
simonhughcom: I think I should go back to using $!.
@Euro20179 seems to think it does not work in dash, but it works with my dash, so it must be a problem on your end.

Euro20179: oh, yes it was a problem with my end.

ytfzf
Euro20179, line 95: what is the point of having `enter,double-click` here?

simonhughcom: do you want me to add a depreciation warning?

ytfzf
```diff
@@ -89,12 +89,16 @@ pid_file="$cache_dir/.pid"
...
 #> config settings
-#list of shortcuts to use in fzf
-#handle_custom_shortcuts will be called if defined (handle_custom_shortcuts can call handle_shortcuts to handle normal shortcuts)
-shortcuts="${shortcuts-alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter}"
+#handle_custom_key_binds will be called if defined (handle_custom_key_binds can call handle_key_binds to handle normal key binds)
+# key_binds sets the bindings to use for fzf
+# any undefined binds will be used to select
+key_binds="${key_binds-alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter,enter,double-click}"
+# these keys allow the use of key_binds without leaving fzf
+persistent_key_binds="${persistent_key_binds-enter,alt-enter,alt-v,alt-d,alt-o,double-click}"
```
Euro20179, line 97: what is the point of `persistent_key_binds` seems a bit redundant?

ytfzf
simonhughcom: For basic selection, as the comment says "any undefined binds will be used to select", i.e. if the bind is not defined in either `handle_key_binds` or `handle_custom_key_binds` then it will be used to select.

ytfzf
simonhughcom: Because you don't want every key bind to be persistent, only some.

Euro20179: keys defined in `persistent_key_binds` seem to do nothing, i commented it out, and they keys did the same thing?

ytfzf
Euro20179: ok, but aren't enter and double-click already used for selection?

ytfzf
simonhughcom: I don't understand, I just did the same and it works perfectly fine?

ytfzf
simonhughcom: Not anymore, they are disabled by default now. This was necessary to make persistent binds work with enter and double-click.
It also has the added benefit of allowing people to choose whatever keys they want for select.

ytfzf
simonhughcom: Try the same experiment, but without looping, or watch if the selection gets reset to the beginning without `persistent_key_binds`

Euro20179: for me it just doesn't keep fzf open, which i'm assuming it's supposed to do. Lemme try without my config file.

simonhughcom: That's odd?
I've tried it with and without my config and it works fine.
What shell are you using?

Euro20179: my /bin/sh is dash

simonhughcom: That's the shell I'm using, just in case I just tried it with bash and it also works there.
It should work.

Are you sure you tried it without your config?

Euro20179: No, no i got it to work.

simonhughcom:
Just tried it with zsh and it does not work there, but that is because of `IFS`, am I correct to assume that zsh is not an official shell we support?

Euro20179:
>  zsh is not an official shell we support?

We could try to get zsh to work, but zsh is a mess, and there seems to be a lot of incompatibility.

simonhughcom:
> > zsh is not an official shell we support?
> 
> We could try to get zsh to work, but zsh is a mess, and there seems to be a lot of incompatibility.

Its fine by me if zsh does not work, I'm just checking that you guys are ok with it.

Euro20179:
> Its fine by me if zsh does not work, I'm just checking that you guys are ok with it.

yes, we are ok with it.

simonhughcom:
Out of the scope of this PR, but it might still be a good idea to have a "official supported shells" section in the readme, just to make sure people don't get confused.

Euro20179:
> but it might still be a good idea to have a "official supported shells" section in the readme

probably.

Euro20179:
```diff
diff --git a/docs/conf.sh b/docs/conf.sh
index fe41dcf..a487586 100644
--- a/docs/conf.sh
+++ b/docs/conf.sh
@@ -286,7 +286,10 @@ sp=""
     # search again
 #in that order, these keys can be changed
 #any keys after will not have default behaviour and the behaviour must be defined in handle_custom_shortcuts
-shortcuts="alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter"
+key_binds="alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter,enter,double-click"
+
+#these are a list of keys that will keep fzf open.
+persistent_key_binds="enter,alt-enter,alt-v,alt-d,alt-o,double-click"
 
 #some helpful variables to keep in mind:
     #selected_key: they shortcut pressed
diff --git a/ytfzf b/ytfzf
index 8c948fe..3db811d 100755
--- a/ytfzf
+++ b/ytfzf
@@ -96,6 +96,8 @@ key_binds="${key_binds-alt-l,alt-t,alt-o,alt-v,alt-d,alt-m,alt-s,alt-enter,enter
 # these keys allow the use of key_binds without leaving fzf
 persistent_key_binds="${persistent_key_binds-enter,alt-enter,alt-v,alt-d,alt-o,double-click}"
 
+[ -n "${shortcuts+shortcuts_is_defined}" ] && printf "\033[33mWarning: shortcuts is defined, use the 'key_binds' variable instead\033[0m\n" >&2
+
 search_prompt=${search_prompt-Search: }
 
 #used when getting the html from youtube
```

This adds a warning for when shortcuts is defined, and also puts `key_binds`, and `persistent_key_binds` in `docs/conf.sh`.
Can you add these things?

(I was going to myself but I am bad at git and honestly just don't want to mess anything up)

simonhughcom:
Do you want me to update `conf.sh`?

Euro20179:
yes

Euro20179:
can you also add this line,
```sh
command -v "handle_custom_shortcuts" > /dev/null 2>&1 && printf "\033[33mWarning: handle_custom_shortcuts is not supported, rename it to handle_custom_key_binds\033[0m\n" >&2
```

PR #325, opened 2021-07-13
feat: video-limit

Euro20179:
-P wont be good w/ thumbnails with this pr but it will work in master.

### DIFF
--- a/docs/conf.sh
+++ b/docs/conf.sh
@@ -193,6 +193,11 @@ preview_side="left"
 #same as --subs=
 sub_link_count=10
 
+#the maximum amount of videos to show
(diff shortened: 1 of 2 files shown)


pystardust:
What about just modifying the get_video_data function to parse only that many entries.
Since we are anyways doing `.[]`, adding a splice there would save the hassle of playing with the video_data

Euro20179:
how would that work?

Euro20179:
Actually I don't think it would work, because it can't operate on each channel individually, it needs to operate after all channels have been scrapped.

pystardust (approved):

PR #328, opened 2021-07-14
feat: persistent_key_binds only with -l

Euro20179:
@simonhughcom, just making sure that this is fine and there isn't a problem with this, if you approve just add a thumbs up emoji to this or something


simonhughcom:
This would likely introduce bugs, ill talk to you about potential solutions later on.

simonhughcom:
So the problem is that some parts of `persistent_key_binds` always have to run, in particular lines 984-994
``` shell
    # if in neither persistent_bind nor key_binds ignore
    # if in persistent_bind, always ignore
    ignore_keys="enter,double-click"
    ignore_binds=
    for key in $ignore_keys; do
        printf "%s%s" "$binds" "$key_binds" | grep -q ",$key," || \
            ignore_binds="${ignore_binds},${key}:ignore"
        printf "%s" "$key_binds" | grep -q ",$key," || continue
        ignore_binds="${ignore_binds},${key}:ignore"
    done
    unset IFS
```
maybe it should be its own separate function.

The easiest way to solve this would be to simply `[ $enable_loop -eq 0 ] && persistent_key_binds=""` and then they won't be set, however I think there is a user case for using persistent key binds without a loop, so maybe introduce a new variable `enable_persistent_key_binds=1` for lines 972-982 and then you can simply `[ $enable_loop -eq 1 ] && enable_persistent_key_binds=1`?

Euro20179:
> So the problem is that some parts of persistent_key_binds always have to run, in particular lines 984-994

why does this have to run? it seemed to work just fine without running.

simonhughcom:
it sets enter and double-click to ignore in fzf, its nothing major but without it they would always be enabled by fzf even if not explicitly set by the user.

Euro20179:
ok, i think i have come up with a good solution.

ytfzf
```diff
@@ -1594,6 +1597,7 @@ parse_opt () {
 
 		l|loop)
 			enable_loop=${optarg:-1}
+			enable_persistent_key_binds=1
```
simonhughcom, line 1600: That should be `enable_persistent_key_binds=${optarg:-1}`

Euro20179: yes

simonhughcom:
Looks good :+1:

PR #334, opened 2021-07-17
Fix: stop removing key binds

simonhughcom:
This PR fixes a bug, it stops persistent key binds from removing keys from `key_binds` which are needed in `handle_key_binds`


Euro20179:
I will review this in a bit.

PR #360, opened 2021-08-13
feat: ytfzf-cur

Euro20179:
If there's a better place to write to $ytfzf_cur, i'll move it there.

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -53,6 +53,12 @@ function_exists "downloader" || downloader () {
 	esac
 }
 
+# directories
+: ${cache_dir=$HOME/.cache/ytfzf}
+
+# files
+: ${ytfzf_cur=$cache_dir/ytfzf_cur}
+
 # format options
 : ${is_audio_only=0}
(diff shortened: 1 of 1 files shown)


pystardust:
what about having the json in the curfile, then the user could have a notification with the thumbnail and all sorts of stuff

ytfzf
```diff
@@ -639,6 +673,9 @@ shift $((OPTIND-1))
 # Main {{{
 json_file=$(mktemp)
 id_file=$(mktemp)
+#provide easy access to the id, and json
+ln -sf "$id_file" "$ytfzf_selected_ids"
+ln -sf "$json_file" "$ytfzf_video_json_file"
```
pystardust, line 678: why not just use this file instead of using the temp file?
as in use the $ytfz_selected_ids file for selected id and so on.
Since we anyways have to mess with pid for both ways

Euro20179: > why not just use this file instead of using the temp file?

fixed

PR #362, opened 2021-08-19
FIX(thumb): Respect thumb method

gvolk0:
No matter what you set it to ytfzf was trying to use the default method (and complaining if it doesnt exist).

This PR fixes that.

My editor also auto-cleaned the trailing whitespace.


Euro20179:
This has already been addressed in the development branch

gvolk0:
😢

PR #366, opened 2021-08-25
feat: -a, -r

Euro20179:


ytfzf
```diff
@@ -485,33 +533,23 @@ interface_text () {
...
 }
 #}}}
 
 # External interface {{{
 interface_external () {
 	video_json_file=$1
-	selected_id_file=$2
-
-	# video_info_text can be set in the conf.sh, if set it will be preferred over the default given below
-	TTY_COLS=$external_menu_len
```
pystardust, line 499: external length variable would be affected here. Maybe its better to have 2 the variables set separately for both cases rather than a function

ytfzf
```diff
@@ -607,8 +644,7 @@ interface_thumbnails () {
 	sort_video_data_fn |
 	fzf -m \
 	--preview "sh $0 -U preview_img '"$thumbnail_viewer"' {} '"$video_json_file"'" \
-	--preview-window "left:50%:wrap" --layout=reverse |
-	trim_id > "$selected_id_file"
```
pystardust, line 611: how would the selection go into the id file here

pystardust: oh i see now, but doing it in the bottom case like this seems like it removes flexibility

pystardust:
also to pick the first few/random video its possible with just jq and the video_json file. jq supports array slices. `.[1:n]`, we could also do random by finding a random number and ` $((rand % n)) `  where `n` is the total number of elements in the array.

ytfzf
Euro20179: > oh i see now, but doing it in the bottom case like this seems like it removes flexibility

I thought about this as well, however if they need to do anything after returning the ids, they can put it in {} and redirect to /dev/null, but i can remove it.

ytfzf
Euro20179: Oh, oops didn't realize this, we're going to keep the function since it reduces a lot of code, the pattern could be like this,

1. call the function
2. set any lengths that weren't set properly

Euro20179:
for picking the first n videos i think sed works better because you can use $ as the last index, but with slices you have to know the length to get the last video.

PR #370, opened 2021-08-29
more readable peertube video duration format

pirate486743186:
changing format from seconds to minutes:seconds for peertube


pystardust:
@pirate486743186  we are rewriting the whole script to make it more modular. Check out the "rewrite" branch. It would be nice if you can contribute directly to that

pirate486743186:
don't over estimate my abilities....
merge anyway

PR #374, opened 2021-08-31
Remove duplicate variables in config file

Mudskipper875:
The `is_download` and `exit_on_opt_error` variables appear twice in the config file.

### DIFF
--- a/docs/conf.sh
+++ b/docs/conf.sh
@@ -147,10 +147,6 @@ thumbnail_quailty=1
 #same as -m
 is_audio_only=0
 
-#download the video instead of watching/listening
-#same as -d
-is_download=0
-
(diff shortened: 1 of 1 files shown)


Euro20179:
For the rewrite, we are planning on putting all configuration options in a man page. That way we can change default settings and not have to worry about people's config as many people just copy the default config.

However we will probably also have a very small sample config that won't be changed. Something like

```sh
#!/bin/sh

#this is a shell script file and can be treated as such.

#by default, scrape subscriptions
scrape="youtube-subscriptions"

#when -D is used this function will be used as the menu.
external_menu () {
    sed 1q
}

```

Mudskipper875:
Oh I see

PR #378, opened 2021-09-01
feat: long_opt_char customization, search-pid template for cache dir

Euro20179:
* able to customize long_opt_char with it's default being -
* change the `cache_dir` template to `search-pid`

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -42,6 +42,10 @@ trim_id () {
 	done
 }
 
+# Traps {{{
+trap 'clean_up' EXIT
+trap 'exit' INT TERM HUP
+#}}}
 
 # }}}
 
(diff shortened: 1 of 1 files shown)


ytfzf
```diff
@@ -756,24 +743,24 @@ parse_opt () {
 		L) info_to_print="$info_to_print,L" ;;
 		c) scrape=$optarg ;;
 		I) info_to_print=$optarg ;;
-		v) # set variable
+		"$long_opt_char") # set variable
```
pystardust, line 746: i don't think this would work, the long_opt_char would be filtered out before parse_opt

Euro20179: there is a check to make shre long_opt_char is not alphanumeric or v, v will be reserved as the only alphanumeric char for a long opt

ytfzf
```diff
@@ -94,11 +98,18 @@ function_exists "downloader" || downloader () {
...
 : ${is_random_select=0}
 
+# option parsing
+: ${long_opt_char:=-}
 
 # scrape
 : ${scrape=youtube}
 : ${sub_link_count=10}
 
+# Option Checks {{{
```
pystardust, line 108: what about a case statement here?
like checking for a `-` or `v`, if anything else then we check for non alphanumberic

Euro20179: we could also add
```sh
[!v-]) print_info "warning: only v and - are reserved for longopts, other chars may be used in the future" ;;
```

to the case statement if we want to use other nonalphanumeric chars but i dont think we should.

pystardust: Yeah sounds good

pystardust (approved):

PR #381, opened 2021-09-02
Multi scrape

Euro20179:
`ytfzf -cO,Y,S search`

the while loop at the bottom could also be,
```sh
for scraper in $(printf '%s' "$scrape" | tr ',' ' '); do
    ...
done
unset scraper
```

but that's not pure sh.

### DIFF
--- a/docs/man/ytfzf.1
+++ b/docs/man/ytfzf.1
@@ -58,14 +58,21 @@ Auto\-play a random result,
 .B \-L
 Print the URL of the selected videos.
 .TP
-.BI \-c " scraper"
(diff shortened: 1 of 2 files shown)


Euro20179:
do not merge

pystardust:
I don't think multi search would be used that much. Multi scraper is pretty cool tho.
We could make it so that a user could implement multi search but adding it to the code feels unnecessary

Euro20179:
fair enough, I've rolled it back.

pystardust (approved):

PR #384, opened 2021-09-03
feat: New options `ytdl-opts` and `ytdl-path`

Mudskipper875:
1. `ytdl-opts`: For passing command-line options to youtube-dl
> ```
> ytfzf -d --ytdl-opts='-o ~/Videos/%(title)s.%(ext)s'
> ytfzf -d --ytdl-opts='--config-location ~/.config/youtube-dl/alt-config'
> ```
2. `ytdl-path`: For using a fork of youtube-dl
> ```
> ytfzf -d --ytdl-path=yt-dlp
> ```

Though these options can already be set by defining  the `downloader` function, this might make it easier for users who aren't familiar with shell scripting.

### DIFF
--- a/docs/man/ytfzf.1
+++ b/docs/man/ytfzf.1
@@ -64,6 +64,15 @@ Auto\-play a random result,
 .B \-L
 Print the URL of the selected videos.
 .TP
+.BI \-\-ytdl\-opts= option
+Pass command\-line options to youtube\-dl when downloading.
+.EX
+.RB "example: " \-\-ytdl\-opts= "\fI\'\-o ~/Videos/%(title)s.%(ext)s\'"
+.EE
+.TP
+.BI \-\-ytdl\-path= path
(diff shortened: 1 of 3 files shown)


Euro20179:
`--ytdl-opts` is nice, but `--ytdl-path` seems redundant because you should have a symlink that leads to it, or you can change the function in the config (when we add it) to use whatever instead.

Although it adds almost no complexity to the program it probably doesnt matter.

Mudskipper875:
> `--ytdl-opts` is nice, but `--ytdl-path` seems redundant because you should have a symlink that leads to it, or you can change the function in the config (when we add it) to use whatever instead.
> 
> Although it adds almost no complexity to the program it probably doesnt matter.

It could also be used to specify the path to an executable which is not in PATH.
```
ytfzf -d --ytdl-path='/path/to/youtube-dl'
```

Euro20179:
> which is not in PATH

true and i think it could be useful if you want to use a custom wrapper script around youtube-dl for some reason

PR #385, opened 2021-09-07
feat: read stdin

Euro20179:
* also adds a search prompt for when no search is given.

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -61,6 +61,8 @@ function_exists "external_menu" || external_menu () {
 #number of columns (characters on a line) the external menu can have
 : ${external_menu_len=210}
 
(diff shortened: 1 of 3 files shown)


ytfzf
```diff
@@ -783,13 +805,24 @@ done
 shift $((OPTIND-1))
 #}}}
 
+# Get search{{{
+search="$*"
+get_search () {
+	case "$search" in
+		-) print_info "Reading stdin\n"; read -r search ;;
+		'') search_prompt_menu ;;
```
pystardust, line 813: What does $input.do here

Euro20179: oops, that line was meant to be removed since it's now redundant to the line `search="$*"`.

fixed.

PR #394, opened 2021-09-19
fix: Remove some subshells for speed

Mudskipper875:


ytfzf
```diff
@@ -654,16 +653,15 @@ interface_thumbnails () {
 	selected_id_file=$2
 
 	# Download thumbnails, only if they're not already downloaded
-	[ -z "$(ls -1 "$thumb_dir" 2>/dev/null)" ] && {
+	find "$thumb_dir" -maxdepth 0 -type d -empty | grep -q . && {
```
Euro20179, line 656: -maxdepth isn't [posix compliant](https://pubs.opengroup.org/onlinepubs/9699919799/)
I had a check that looked like this, but avoided it because of globbing and possible weird behavior.
```sh
set +f
ls "$thumb_dir"/* > /dev/null 2>&1 || { ...
```

Mudskipper875: `-maxdepth` can be removed in this case and we will still have the same result, unless we expect to have have another empty directory inside `$thumb_dir`.
If we grep the name of the directory it will possibly solve even that issue:

```shell
find "$thumb_dir" -type d -empty | grep -q "$thumb_dir" && {
```
If this is okay, I'll push it

Euro20179: Yes this is good.

Mudskipper875:
I have a another suggestion. Can the temp files be moved into the cache directory instead of having separate cache and temp directories?

Euro20179:
> I have a another suggestion. Can the temp files be moved into the cache directory instead of having separate cache and temp directories?

We could but, the files in the cache directory are easy to interact with, I think it would be better to keep them separate, we could have `$session_cache_dir/tmp`

pystardust:
I think its a good idea to have the temp files in tmp/ inside session cache, this also avoids `mktemp` which isn't POSIX.

Euro20179:
Ngl, having tmp files with proper names, and not removing them until the program dies is actually quite nice.

Mudskipper875:
I just learnt that piping output to a command also creates a subshell. So two of these commits: add90a1 and 8e0f6c9 probably don't make a difference in performance. 😅

Euro20179:
> I just learnt that piping output to a command also creates a subshell.

I thought this was the case lol, all good though.
Also pipes usually look cleaner anyway.

PR #397, opened 2021-09-25
History

Euro20179:


ytfzf
```diff
@@ -527,12 +563,13 @@ thumbnail_video_info_text () {
 interface_scripting () {
 	video_json_file=$1
 	selected_id_file=$2
+	data_string_template='.[]|"\(.title)\t|\(.channel)\t|\(.duration)\t|\(.views)\t|\(.date)\t|\(.ID)"'
 	case 1 in
 		"$is_auto_select")
-		jq -r '.[]|.ID' < "$video_json_file"  | head -n "$scripting_video_count" > "$selected_id_file"
+		jq -r "$data_string_template"  < "$video_json_file"  | head -n "$scripting_video_count" |  trim_id  | > "$selected_id_file"
```
pystardust, line 569: why do we need title channel etc here?

ytfzf
```diff
@@ -427,7 +435,7 @@ scrape_peertube () {
 				channel: .channel.displayName,
 				thumbs: .thumbnailUrl,
 				duration: "\(.duration / 60 | floor):\(.duration % 60)",
-				views: .views,
+				views: "\(.views)",
```
pystardust, line 438: seen this in other places also, isn't `.views` equivalent to `\(.views)`?

pystardust (approved):

ytfzf
Euro20179: so it gets logged in the history file correctly.

ytfzf
Euro20179: "\(.views)" is a string, so it can be added to other strings, it throws an error when views is an int.

PR #400, opened 2021-09-28
Use invidious

Euro20179:


Mudskipper875:
I'm getting low quality thumbnails after merging this branch. Also, is there an option to revert back to using youtube?

Euro20179:
you can set `yt_thumbnail_quality` to high or maxres, currently no there is no youtube option but i cojld add it back, invidukus seems better, for subscriptions it defaults to youtube because its more stable for scraping a lot

Mudskipper875:
Oh okay got it 👍

PR #406, opened 2021-10-01
Read the search-query from stdin even without specifying '-'

Mudskipper875:
Seemed a little complicated to use two different functions to process the search-query

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -846,7 +846,10 @@ shift $((OPTIND-1))
 #}}}
 
 # Get search{{{
+set -f
+set -- $*
 search="$*"
+set +f
 [ -z "$search" ] || [ "$search" = '-' ] && search_prompt_menu
 #}}}
 
--- a/ytfzf
+++ b/ytfzf
@@ -65,13 +65,6 @@ trim_url () {
 	done
 }
 
-search_is_empty () {
-	case "$search" in
-		-|'') return 0 ;;
-	esac
-	return 1
-}
-
 # Traps {{{
 [ -z "$UEBERZUG_FIFO" ] && trap 'clean_up' EXIT
 [ -z "$UEBERZUG_FIFO" ] && trap 'exit' INT TERM HUP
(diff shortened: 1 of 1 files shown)


ytfzf
```diff
@@ -104,8 +97,9 @@ function_exists "search_prompt_menu" || search_prompt_menu () {
 	if [ $is_ext_menu -eq 1 ]; then
 		search="$(printf '' | external_menu)"
 	else
-		printf "Search: "
+		printf "Search\n> "
 		read -r search
+		printf '\e[1A\e[K\r%s\n' "> $search"
```
Euro20179, line 102: \e isn't really standard, use \033 instead, also can you like explain what this those escape sequences do.

Mudskipper875: Those escape sequences erase the previous line and return the cursor to the beginning of the line.

`printf '\033[1A\e[K\r%s\n' "> $search"` will erase the previous line and print '> search-query'

1. This is the output if the search query is given as an argument.
```shell
$ ytfzf 'Search query'
Scraping YouTube
Fetching thumbnails...
```

2. This is the output when you are prompted to type the search query after invoking ytfzf.
```shell
$ ytfzf
Search
> 
```
```shell
$ ytfzf
Search
> Search query
Scraping YouTube
Fetching thumbnails...
```

3. This is the output when the search query is piped to ytfzf. This is why I changed the search prompt have 2 lines because if the search query is piped to stdin, then the shell prompt gets deleted if there is only a one line prompt.
```shell
$ echo 'Search query' | ytfzf
> Search query
Scraping YouTube
Fetching thumbnails...
```
`echo 'Search query' | ytfzf -` also works still.

ytfzf
```diff
@@ -104,8 +97,9 @@ function_exists "search_prompt_menu" || search_prompt_menu () {
 	if [ $is_ext_menu -eq 1 ]; then
 		search="$(printf '' | external_menu)"
 	else
-		printf "Search: "
+		printf "Search\n> "
 		read -r search
+		printf '\033[1A\e[K\r%s\n' "> $search"
```
Euro20179, line 102: You forgot to change the second \e

Mudskipper875: Oh damn.. 😅

Issue #243, opened 2021-05-14
support notify download success or fail


v6cc:
I have a unstable net environment , always download fail,  could you support use `notify-send`  or something ,  notify download success or fail?

Euro20179:
After way too long, it notify-sends failure if download fails.

It already notified when it was successful.

Failure notification is in the development branch, and will be in 2.6.0.

PR #409, opened 2021-10-03
Notification upon download complete

kyteinsky:
Added notification popup (notify-send) to video/audio downloads to notify the user of the status of the download (successful or failed).

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -1067,6 +1067,13 @@ open_player () {
 		else
 			youtube-dl -f "$video_pref"  "$@"  $YTFZF_SUBT_NAME || video_pref= open_player "$@"
 		fi
+
+    if [ $? -eq 0 ]; then
+        dl_status="Successful"
+    else
+        dl_status="Failed"
+    fi
+    notify-send "[ytfzf] Download $dl_status!" "$*"
 	fi
 }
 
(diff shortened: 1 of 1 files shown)


kyteinsky:
There is an error from youtube-dl that pops up even upon successful download:
```
ERROR: '' is not a valid URL. Set --default-search "ytsearch" (or run  youtube-dl "ytsearch:" ) to search YouTube
```
But this does not concern this pull request much other than a download failed notification even upon successful download. This should be solved when that issue is solved in the youtube-dl repo.

kyteinsky:
Solves #243

ytfzf
```diff
@@ -1067,6 +1067,13 @@ open_player () {
 		else
 			youtube-dl -f "$video_pref"  "$@"  $YTFZF_SUBT_NAME || video_pref= open_player "$@"
 		fi
+
+    if [ $? -eq 0 ]; then
+        dl_status="Successful"
+    else
+        dl_status="Failed"
+    fi
+    notify-send "[ytfzf] Download $dl_status!" "$*"
```
Euro20179, line 1076: This should probably be after line 1068 so that $? refers to the right command.

kyteinsky: Oh I didn't know that could be a problem. I just wanted to save some lines of code.

PR #424, opened 2021-11-16
New Feature: change country for trending and searches

RaviRahar:
Allows to change country with ` --country ` flag. Example:
` ytfzf --trending=music --country=US `

Can also be set to a default by exporting ` YTFZF_CON ` variable.


RaviRahar:
Nothing drastic

Euro20179:
This has been added with the --region  option

PR #430, opened 2021-12-10
update the installation for other distros

mazylol:


mazylol:
updated dependencies in the install readme

Euro20179:
In the rewrite I updated it to this

1. Install the dependancies listed [above](#Dependencies)
2. Run the following commands
```sh
git clone https://github.com/pystardust/ytfzf
cd ytfzf
sudo make
```
* ytfzf may be in some repositories, such as the AUR, however the above way is the recommended way to install.


Because I agree that the current install page is a bit excessive.

PR #432, opened 2021-12-16
add mpv --ytdl option for audio/video player

rumly111:
I usually put `ytdl=no` in my mpv config, just to speed up the startup time. But that breaks ytfzf. Adding mpv --ytdl to explicitly tell mpv use youtube_dl solves the problem.


Euro20179:
I don't think this needs to be applied for everyone, if this is just for your situation.

Issue #64, opened 2021-03-03
Thumbnails non-functional in Wayland using normal terminal emulators


littlediobolic:
Thumbnails do not display in Wayland using a normal terminal emulator due to ueberzug relying on X. This is obviously an upstream issue but something I feel is worth mentioning. For example gnome-terminal does not work.

I tested in [kitty](https://github.com/kovidgoyal/kitty)however,  which has its own image protocol and it does work, but in 99% of terminals it will likely not work by default.

duffydack:
Nice to know kitty works, I use kitty with ranger.

SamsiFPV:
I did a little bit of testing on this. It works on X11 based terminal emulators such as urxvt. Wayland based ones such as terminator don't work. Which has definitely to do with ueberzug being X11 based. Adding an additional option for us Wayland users would be appreciated.
PS: @littlediobolic i didn't got it to work with kitty, how did you do that? It's a wayland native app so it shouldn't work, since ytfzf only supports ueberzug and not kittys own protocol.

napcok:
@SamsiFPV Terminator is not Wayland based, and thumbnails works very well in Terminator on X11.

SamsiFPV:
@napcok wrong wording form my side, i meant it runs wayland natively. My bad.

pystardust:
How about we use an external viewer? 
 I was able to make it work with sxiv (any viewer that auto updates when the file changes) . The only caveat is that it wouldn't look great since you have 2 windows, one with names and the other with thumbnails.

arthureroberer:
maybe a pretext to re-invent ueberzug but for wayland?
could be useful for tons of things

littlediobolic:
> PS: @littlediobolic i didn't got it to work with kitty, how did you do that? It's a wayland native app so it shouldn't work, since ytfzf only supports ueberzug and not kittys own protocol.

@SamsiFPV I did not do anything special to get it working. I just downloaded kitty, ran it, and thumbnails displayed properly. I can send logs if necessary but I am unsure what caused it to work.

tgunnoe:
I would recommend looking at kitty https://sw.kovidgoyal.net/kitty/graphics-protocol.html

DanKGooGLy:
What about sixel?

Euro20179:
[...]

PR #437, opened 2021-12-30
feat: add kitty/icat thumbnail viewer support

ionenwks:
Tried this and it works great on both x11 and wayland (it's non-generic but it's still something for #64)

Prefer this over ueberzug on x11 as well given the image will work properly when switching terminal tabs.

I've seen mentions of using `transfer-mode file` with fzf and not having a separate --clear, but it displayed nothing if I did either of these. Not that I know kitty/icat that well so there may be something to improve.

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -966,7 +966,7 @@ preview_start () {
 			ueberzug layer --parser simple < "$UEBERZUG_FIFO" 2>/dev/null &
 			exec 3> "$UEBERZUG_FIFO" # to keep the fifo open
 			;;
-		chafa|chafa-16|chafa-tty|catimg|catimg-256|display|w3m) : ;;
+		chafa|chafa-16|chafa-tty|catimg|catimg-256|display|kitty|w3m) : ;;
 		*)
 			"$thumbnail_viewer" "start" "$FZF_PREVIEW_COLUMNS" "$FZF_PREVIEW_LINES" 2>/dev/null ;;
 	esac
@@ -977,7 +977,7 @@ preview_stop () {
 		ueberzug)
 			exec 3>&- # close file descriptor 3, closing ueberzug
 			;;
-		chafa|chafa-16|chafa-tty|catimg|catimg-256|display|w3m) : ;;
+		chafa|chafa-16|chafa-tty|catimg|catimg-256|display|kitty|w3m) : ;;
 		*)
 			"$thumbnail_viewer" "stop" "$FZF_PREVIEW_COLUMNS" "$FZF_PREVIEW_LINES" 2>/dev/null ;;
 	esac
@@ -1062,6 +1062,13 @@ preview_display_image () {
 			dep_check "display" || die 3 "\nimagemagick is not installed\n"
 			killall display
 			display "$thumb_dir/${id}.jpg" ;;
+		kitty)
+			printf '\n'
+			[ "$TERM" = "xterm-kitty" ] || die 3 "\nnot running inside kitty\n"
+			dep_check "kitty" || die 3 "\nkitty is not installed\n"
+			get_ueberzug_positioning "$FZF_PREVIEW_COLUMNS" "$FZF_PREVIEW_LINES" "$fzf_preview_side"
+			kitty +kitten icat --clear --transfer-mode stream
+			kitty +kitten icat --place "${width}x${height}@${x}x${y}" --scale-up --transfer-mode stream "$thumb_dir/${id}.jpg" ;;
 		w3m)
 			while true; do
 				printf "%b\n%s;\n" "0;1;10;130;$((FZF_PREVIEW_COLUMNS*5));$((FZF_PREVIEW_COLUMNS*3));;;;;$thumb_dir/${id}.jpg" 3 | $w3mimgdisplay_path
--- a/ytfzf
+++ b/ytfzf
@@ -1063,8 +1063,6 @@ preview_display_image () {
 			killall display
 			display "$thumb_dir/${id}.jpg" ;;
 		kitty)
-			printf '\n'
-			[ "$TERM" = "xterm-kitty" ] || die 3 "\nnot running inside kitty\n"
 			dep_check "kitty" || die 3 "\nkitty is not installed\n"
 			get_ueberzug_positioning "$FZF_PREVIEW_COLUMNS" "$FZF_PREVIEW_LINES" "$fzf_preview_side"
 			kitty +kitten icat --clear --transfer-mode stream
--- a/docs/man/ytfzf.1
+++ b/docs/man/ytfzf.1
@@ -141,6 +141,9 @@ Imagemagick's display module.
 .IR w3m
 Uses a workaround to get w3m to work in fzf, may take up a lot of cpu.
 .TP
+.IR kitty
+uses kitty's builtin icat to display an image.
+.TP
 .IR <custom>
 Additional viewers can be put into $YTFZF_THUMBNAIL_VIEWERS_DIR.
 .RE
--- a/docs/man/ytfzf.5
+++ b/docs/man/ytfzf.5
@@ -211,6 +211,9 @@ Uses imagemagick's preview image
 Uses a workaround to get w3m to work in fzf, may take up a lot of cpu,
 make sure $w3mimgdisplay_path is set to the path to w3mimgdisplay
 .TP
+.IR kitty
+Uses kitty's builtin icat to display an image.
+.TP
 .IR custom
 Calls the user defined img_display_function()
 .TP


ytfzf
```diff
@@ -1062,6 +1062,13 @@ preview_display_image () {
 			dep_check "display" || die 3 "\nimagemagick is not installed\n"
 			killall display
 			display "$thumb_dir/${id}.jpg" ;;
+		kitty)
+			printf '\n'
+			[ "$TERM" = "xterm-kitty" ] || die 3 "\nnot running inside kitty\n"
```
Euro20179, line 1067: This is not a good way to check if it's running in kitty, because $TERM is just what the user puts in `.bashrc` not necessarily the terminal they're running

ytfzf
```diff
@@ -1062,6 +1062,13 @@ preview_display_image () {
 			dep_check "display" || die 3 "\nimagemagick is not installed\n"
 			killall display
 			display "$thumb_dir/${id}.jpg" ;;
+		kitty)
+			printf '\n'
```
Euro20179, line 1066: Since kitty takes a position to put the image, we don't need this.
Things like chafa print actual text, so we need to print a new line before printing the image.

Euro20179:
Gj on getting this to work btw, i've never gotten it to work.

ytfzf
Euro20179: You could try something like
```sh
kitty +icat --detect-support --detection-timeout=4 || die 3 "\nnot running inside kitty\n"
```
and also put that after the `dep_check kitty`
You can also change detection-timeout to whatever you think is good (the default is 10 seconds which seems long)

ionenwks: Yes, any use of --detect-support or even --detection-timeout seem to break things.

I've tried to do other things like `kitty +kitten query_terminal --wait-for=2 name` but that /also/ breaks it. I think kitty misdetect things (thus the need for transfer-mode) and about anything will break it.

Don't really have any better ideas right now, TERM was just a basic check in case "oops happened to start it in the wrong terminal" so things don't get crazy (not that you can't just ctrl+c out of it but well).

ionenwks: ...but well, guess users may also want to set TERM to something else and while keeping this working, so removing the TERM check may be better. It'd just rely on users using the right settings.

ytfzf
ionenwks: Right, hadn't really thought about it.

ytfzf
Euro20179: aight here's a hacky solution
in `preview_start` we can add a case for kitty
```sh
kitty)
    kitty +kitten query_terminal name &
    pid="$!"
    sleep 1
    kill "$pid" && die 3 "not using kitty terminal"
```

we can change `sleep 1` to be shorter but we can't make it too short.

---

This works because when the terminal isn't kitty, `query_terminal` takes AGES, so if we try killing it before it's done kill will be sucessful meaning we're not in kitty, if kill is successful we are in kitty.

I feel like we could do better than this though.

Euro20179: > It'd just rely on users using the right settings.

This is how i'd do it too.

David-Else:
It doesn't seem to work :( I am using the latest Kitty.
```
$ ytfzf --version
ytfzf: 2.1.rc-4 
```
```
ytfzf -t --thumb-viewer=kitty
```
```
Failed to launch child: stop
With error: No such file or directory
Press Enter to exit.
```
It then launches into `vi`!?
Screenshot from 2022-02-03 18-06-14

If I quit `vi` it shows:

Screenshot from 2022-02-03 18-06-35

and if I select another it repeats the crash sequence.

I did have to uninstall and re-install `ytfzf` recently, I have not tried to use kitty before. `ytfzf` is working otherwise.

Can anyone help?

ionenwks:
While it mostly works, the support was removed due to #454 and by doing this it's probably trying to run kitty itself in the terminal and doing weird things.

Fortunately, the issue led to the addition of a way to set it up easily anyway (for me it always works fine, so I use it), e.g. me  I have in ~/.config/ytfzf/conf.sh:
```
thumbnail_viewer=${YTFZF_THUMBNAIL_VIEWERS_DIR}/kitty-viewer
```
And in the executable ~/.config/ytfzf/thumbnail-viewers/kitty-viewer
```bash
#!/usr/bin/env bash
[[ ${1} == view ]] || exit 0

kitty +kitten icat --clear --transfer-mode file

[[ ! -e ${2} ]] ||
	kitty +kitten icat --place "${5}x${6}@${3}x${4}" --scale-up --transfer-mode file "${2}"
```
(forgive the intentional bash, I don't do posix sh, see #454 for the other example)

Euro20179:
Try using the add-on in the adding folder by putting it in `~/.config/ytfzf/thumbnail-viewers/kitty-viewer` (make sure it's named kitty-viewer, not kitty), then `chmod +x ~/.config/ytfzf/thumbnail-viewers/kitty-viewer`, then you can run `ytfzf --thumb-viewer=kitty-viewer` or add `thumbnail_viewer=kitty-viewer` to your config file.

If this doesn't work, then you are like me in that the kitty viewer is extremely buggy lol.

I've also recently discovered that using the command `kitty +kitten icat --silent --clear --transfer-mode file`, instead of `kitty +kitten cat --clear --transfer-mode file` reduces the amount of errors. (I'll probably put this in the add-on at some point)

ionenwks:
> EDIT: for my sanity please change `[[ ! -e ${2} ]] ||` to `[[ -e ${2} ]] &&` LOL.

There is a difference between the two, the latter ends with $? = 1 if the test failed. Meaning kitty-viewer exits with an error (not that it matters here), and it'd also prevent usage of `set -e` if wanted to.

Sometime I do things like `! condition || dostuff || die "something went wrong!"` which is different than `condition && dostuff || die` given it'll die if the condition failed

Euro20179:
> There is a difference between the two, the latter ends with $? = 1 if the test failed. Meaning kitty-viewer exits with an error (not that it matters here), and it'd also prevent usage of set -e if wanted to.

> Sometime I do things like ! condition || dostuff || die "something went wrong!" which is different than condition && dostuff || die given it'll die if the condition failed

Ohh, yes this makes sense. (I also like using `||` and `&&` more than if then, but in these situations I'd probably use if then lol)

David-Else:
I added the config file `~/.config/ytfzf/thumbnail-viewers/kitty-viewer`. It works if I:
```
ytfzf -t --thumb-viewer=kitty-viewer
```
but with:
```
thumbnail_viewer=${YTFZF_THUMBNAIL_VIEWERS_DIR}/kitty-viewer
ytfzf -t
```
It does not seem to accept the env variable and goes to the default viewer which does nothing.

The thumbnails are in ulra-low resolution and barely viewable, how can I increase it?

Could someone please make a pull request and get this functionality back into `ytfzf` so `--thumb-viewer=kitty` works?

Cheers, thanks for the help!

Euro20179:
> `thumbnail_viewer=${YTFZF_THUMBNAIL_VIEWERS_DIR}/kitty-viewer`

you have to put this in your `~/.config/ytfzf/conf.sh` file.

> The thumbnails are in ulra-low resolution and barely viewable, how can I increase it?

`--thumbnail-quality=maxres` or put `thumbnail_quality=maxres` in your config.

I'll probably change the default of this option since most people have good internet.

> Could someone please make a pull request and get this functionality back into `ytfzf` so `--thumb-viewer=kitty` works?

Maybe, for now it'll stay an addon, but I will rename the addon to kitty-viewer, because otherwise it doesn't work.

David-Else:
@Euro20179 Thanks! 

I don't know how things work here, where do I find addons, and is there an official way to install them? Do I need to return to this thread to find out about updates?

ionenwks:
+1 for higher res default

> Maybe, for now it'll stay an addon, but I will rename the addon to kitty-viewer, because otherwise it doesn't work.

Haven't looked in-depth but $thumbnail_viewer seems to be launching kitty directly rather than the viewer at some place, `kitty` works fine with my conf.sh given I use the full path and that doesn't mix up.

Euro20179:
reply to @David-Else,

> where do I find addons

There's an addons folder in this repository, with all the different types of addons, you can find it [here](https://github.com/pystardust/ytfzf/tree/master/addons)

There's also some instructions in the readme under the `Install` section how to install and use them.

> and is there an official way to install them?

Currently no, however i am planning on adding a `--install-addons` option, so when you clone the repository, you can also do `./ytfzf --install-addons` and it'll install all the addons.

>  Do I need to return to this thread to find out about updates?

Yeah, there's a releases link on the left side of the page, i'd check there. `ytfzf` is in a couple of repositories, so if you're lucky you can use that. The only one i maintain is `ytfzf-git` in the aur.
TBH, I should probably announce in the discord when there's an update lol. I could start doing that next update.

Euro20179:
reply to @ionenwks 

> Haven't looked in-depth but $thumbnail_viewer seems to be launching kitty directly rather than the viewer at some place, kitty works fine with my conf.sh given I use the full path and that doesn't mix up.

Yes, if you name the viewer `kitty` and use `ytfzf --thumb-viewer=kitty` it'll use the terminal, this is because ytfzf does just a simple check to see if it can run `kitty`, if it can it usees that as the viewer, otherwise it uses `${YTFZF_THUMBNAIL_VIEWERS_DIR}/kitty`, that's why i named it `kitty-viewer`

Euro20179:
The default thumbnail quality is now `high` (355b29b4c69213347144c1af77c47b8da8a0524f)

David-Else:
> The default thumbnail quality is now `high` ([355b29b](https://github.com/pystardust/ytfzf/commit/355b29b4c69213347144c1af77c47b8da8a0524f))

Thanks for all the updates that are appearing, I just cloned the lastest:
```
$ ytfzf --version
ytfzf: 2.2.alpha-16 
```
but now the thumbnails don't display anymore, it is just:
```
ytfzf -t --thumb-viewer=kitty-viewer
```
Screenshot from 2022-02-03 21-51-01

David-Else:
@Euro20179 I was just trying the new install addons, but the code does not create the directories...
```
./ytfzf --install-addons
installing thumbnail-viewers
cp: cannot create regular file '/home/david/.config/ytfzf/thumbnail-viewers': No such file or directory
installing interfaces
cp: cannot create regular file '/home/david/.config/ytfzf/interfaces': No such file or directory
installing scrapers
cp: target '/home/david/.config/ytfzf/scrapers' is not a directory
installing sort-names
cp: target '/home/david/.config/ytfzf/sort-names' is not a directory
installing url-handlers
cp: cannot create regular file '/home/david/.config/ytfzf/url-handlers': No such file or directory
```

Euro20179:
@David-Else 
fixed (a257688b92f217807a09a0d574bf1e5e9247eff1)

David-Else:
@Euro20179 Cheers, I reinstalled and run `./ytfzf --install-addons`, they now all install OK :)

New problem, all these files and directories exist but it says they do not?!

```
ytfzf --version
ytfzf: 2.2.alpha-16 

ytfzf -t --thumb-viewer=kitty-viewer
```

Screenshot from 2022-02-04 09-01-08

Euro20179:
It probably doesn't have execute permissions, `chmod +x ~/.config/ytfzf/thumbnail-viewers/kitty-viewer`

Euro20179:
@David-Else hopefully [this commit](1a7ea13c422673396fe39d3dc174e50334ab88bb) fixed it.

David-Else:
@Euro20179 Afraid it is not working, I cloned and installed it all again. See the images, it says the file does not exist, but it does, and it displays fine on the command line:
Screenshot from 2022-02-04 15-47-36
Screenshot from 2022-02-04 15-47-33

Euro20179:
what is the output of `ls -l /bin/sh`,

I see what the problem is, and i'm fairly certain what the cause is but i'm not sure how to fix it since i can't reproduce this.

David-Else:
@Euro20179 I am using Rocky Linux 8 (same as RHEL/Centos):
```
ls -l /bin/sh
lrwxrwxrwx. 1 root root 4 Oct  9 12:06 /bin/sh -> bash
```

Euro20179:
Im not home right now, i wont be for severl hours, so for right now all i can suggest is installing the dash shell, and running `dash ytfzf ...`

Euro20179:
@David-Else
You could change line `1314` to 
```
IFS='`'
```
if i understand the problem correctly.

David-Else:
> IFS='`'

I tried, afraid that does nothing :(

Euro20179:
Ok, another idea, add `unset IFS` before line 1302 (`set --`)
then on line 1302 change it to `IFS="${tab_space}"`

David-Else:
@Euro20179 Sorry, that does not fix it. Thanks again for looking into this!

Euro20179:
I'm not sure what the problem is, so i would say to download different commits until you find one that works.
[Here is the list of commits](https://github.com/pystardust/ytfzf/commits/master)
To download a commit, clone ytfzf, run `git checkout "commit number here"`.
eg:
```sh
git checkout "5281b9c4eea296b24183405a8056251559889947"
./ytfzf --thumb-viewer=kitty-viewer ...
```
(You dont have to reclone for each commit)
A good starting commit would be [this one](https://github.com/pystardust/ytfzf/commit/d7ba845b5c054aa512ae79a8cb250305ab6fcb03).

If you find one that works tell me which commit works.

David-Else:
I have tried a few and gave up, I don't know what I am doing. I will give up, but if it works some day plz post here and I will try again!

Euro20179:
it works for me and i cant reproduce this which makes it hard to fix.

Euro20179:
@David-Else 
made some changes to the relevant code, might be worth updating and trying again.

David-Else:
@Euro20179 Sorry, no change :(
```
$ ytfzf --version
ytfzf: 2.2.alpha-22
```
I did install addons after re-cloning

Screenshot from 2022-02-05 09-00-44

ionenwks:
Just gave it a try as well, broken for me as well but works fine if I revert commit 0cffef397d830b03acddad6bbdca9eda04ac0105

By the way, will there be a way for system install of add-ons? e.g. reading from a build-time specified path like /usr/share/ytfzf (after checking user's ~/ first). For distros packaging this it makes more sense.

ionenwks:
> Just gave it a try, broken for me as well but works fine if I revert commit [0cffef3](https://github.com/pystardust/ytfzf/commit/0cffef397d830b03acddad6bbdca9eda04ac0105)

On an extra note, using at commit 66e6ab9bc521cde596a5a29fced2d8754a32d9c3 was broken for me too, so only using at commit afac154f84b8addad73a91f0bf07830865042e29 works for me in recent history.

Euro20179:
> By the way, will there be a way for system install of add-ons by next release? e.g. reading from a build-time specified path like /usr/share/ytfzf (after checking user's ~/ first). For distros packaging this it makes more sense.

Sounds like a good idea since this also solves my problem of wanting to put `make addons` in the Makefile.

I'm gonna try implementing thumbnails a completely different way and see what happens, cause whatever breaking this is super arbitrary.

Euro20179:
@ionenwks @David-Else 

new commit  (1ce2382c52c1083234625013f695fa3906a0f3e5) (2.2.alpha-23)

I think this should work since it avoids messing with IFS and using super technical features of shell, but we'll see.

Euro20179:
> 

    By the way, will there be a way for system install of add-ons by next release? e.g. reading from a build-time specified path like /usr/share/ytfzf (after checking user's ~/ first). For distros packaging this it makes more sense.

> Sounds like a good idea since this also solves my problem of wanting to put make addons in the Makefile.

After some fiddling, this will be super messy especially if you want a build time set path, so i'll probably stick with `--install-addons` for now.

ionenwks:
> After some fiddling, this will be super messy especially if you want a build time set path, so i'll probably stick with `--install-addons` for now.

You could sed ytfzf on install to change the path, or just default to /usr/share/ytfzf then packagers can patch/sed that path themselves (I have to do this often as a packager and I don't mind it). But if there's no support at all for a system path then it'd be messier.

I do something similar for my own [bash-based project](https://github.com/ionenwks/iwdevtools) which shares sourced includes from lib/ in multiple scripts installed to /usr/share (by default it tries to find them in the cloned source, and if installed then it uses the system path).

> new commit (1ce2382) (2.2.alpha-23)
>
> I think this should work since it avoids messing with IFS and using super technical features of shell, but we'll see.

Seems to work for me (albeit given current kitty-viewer addon transfer-mode is still stream on 2nd line I get some misplaced thumbnails again, `file` eliminates these at least for me).

Euro20179:
> or just default to /usr/share/ytfzf 

I could do that., but couldn't package maintainers also check if $YTFZF_CUSTOM_SCRAPERS_DIR is set for example, and  if not default to ~/.config/ytfzf/scrapers?

David-Else:
@Euro20179 **YOU DID IT!!** It works :)

There are a couple of little bugs. The image seems to randomly move from the right to the left. Any image can swap at any time:

Screenshot from 2022-02-05 16-12-25
Screenshot from 2022-02-05 16-12-36
Screenshot from 2022-02-05 16-12-41

Also there is a problem with `curl`:
```
Fetching thumbnails...
curl: option -fLZ: is unknown
```
```
$ curl --version
curl 7.61.1 (x86_64-redhat-linux-gnu) libcurl/7.61.1 OpenSSL/1.1.1k zlib/1.2.11 brotli/1.0.6 libidn2/2.2.0 libpsl/0.20.2 (+libidn2/2.2.0) libssh/0.9.4/openssl/zlib nghttp2/1.33.0
Release-Date: 2018-09-05
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp 
```

Euro20179:
i cant fix the image moving which is why kitty is staying as an addon
the curl thing is also not a bug, we run 2 curl commands, if the first fails with exit code 2 you have an old version of curl, and itll use a backup command.

ionenwks:
For the image thing, try to change `stream` to `file` in kitty-viewer, for me it never happens anymore with that but mileage may vary given it's based on cpu speed as far a I can tell (some kinda race I think).

Optimal would be to do as kitty upstream suggest and not attempt to use icat for integration but rather write our own implementation that match our use case but I think making it work properly with fzf may be a bit beyond the scope of this project.

Euro20179:
> I think making it work properly with fzf may be a bit beyond the scope of this project.

Ive thought about writing a custom image viewer for this project, problem is all image viewers actually use escape - codes that need to be interpreted by the terminal.

The exeptions are ueberzug (it spawns its own window), and w3m (idk how w3m does it but im fairly certain its not escape sequences)

Euro20179:
> For the image thing, try to change stream to file in kitty-viewer, for me it never happens anymore with that but mileage may vary given it's based on cpu speed as far a I can tell (some kinda race I think).

Wait what, i tried this before and it broke fzf, so i thought it didn't work, I just tried it again and it works flawlessly, i'll be updating the addon.

David-Else:
YES!!! It now works 100%, THANKS everyone :)

ionenwks:
> Wait what, i tried this before and it broke fzf, so i thought it didn't work, I just tried it again and it works flawlessly, i'll be updating the addon.

It has a tendency to misbehave in chroot or remotely, or perhaps some other factors? I did have issue with `file` originally which made me use `stream` but then it just worked fine when I tried it again and never went back.

> YES!!! It now works 100%, THANKS everyone :)

:+1:

Euro20179:
Im gonna try kitty out for a while, if it works this good always ill put it back into the main scriot, cause in this state its almost better than ueberzug, its just a bit slow.

ionenwks:
> Im gonna try kitty out for a while, if it works this good always ill put it back into the main scriot, cause in this state its almost better than ueberzug, its just a bit slow.

Yeah it's the --clear command that does that, it feels seamless without it but then the image overlaps badly when there's a change in aspect ratio (especially noticeable on odysee which is less consistent). A potential workaround would be to use imagemagick's convert to resize adding transparent borders (you can even pipe that into icat without a temporary file) but it was still kind of messy when I tried it and decided to stick with --clear and haven't tried more.

Euro20179:
>  A potential workaround would be to use imagemagick's convert to resize adding transparent borders

I can fiddle around with it for a bit, if you find a good solution i'll try it.

PR #444, opened 2022-01-04
fix Makefile to make more MacOS friendly

aberezin:
Moving man pages to /usr/local/share instead of /usr/share for MacOS bc /usr is part of a readonly disk in Catalina


Euro20179:
addressed in #445

PR #452, opened 2022-01-06
Fix error-handing and options of curl in interface_thumbnails()

soystemd:


ytfzf
```diff
@@ -1160,7 +1160,8 @@ interface_thumbnails () {
 			output = \"'"$thumb_dir"'/\(.ID).jpg\""' \
 				< "$video_json_file" > "$curl_config_file"
 			# curl version > 7.66.0 ( else remove -Z )
-			curl -Z -K "$curl_config_file" || curl -K "$curl_config_file"
+			curl -fLZ -K "$curl_config_file"
+			[ $? -eq 2 ] && curl -fLZ -K "$curl_config_file"
```
Euro20179, line 1164: I think `-Z` is supposed to be removed from this curl call.

soystemd: Yeah my bad, I just fixed it.

PR #495, opened 2022-02-26
Add mpvq addon to queue videos in mpv.

jose1711:


Euro20179:
This actually helped me find something I don't like,

You had to use `/tmp` because `$session_temp_dir` is not set in url-handlers, I will fix this.

But this is good for now

PR #515, opened 2022-03-27
feat: ard scraper

Xystel:
This is a scraper for the media library from the german broadcast service [ARD](https://www.ardmediathek.de/).

# Features
- scrapes the API from [ARD](https://www.ardmediathek.de/) and get the video link
- thumbnail fetching
- respects thumbnail resolution choice
- `--pages` flag is supported. It is implemented as a multiplication of the default page size. This avoids making multiple connections.

# Not supported
- `alt-p` to search for next page

# Bugs
With the default video player, it opens only the audio. With following function in `conf.sh` it opens the video for me
```sh
video_player () {
	mpv "$@"
}
```

# Examples
Basic example:
```sh
ytfzf -c ard tagesschau
```

Example with amount of pages:
```sh
ytfzf -c ard --pages=3 sport
```

# Credits
The `pad_left` function is copied from the [source code of ytfzf](https://github.com/pystardust/ytfzf/blob/9f63601e23de30d15a2337b33cac0ecf9a2bfabf/ytfzf#L678-L681). I didn't mentioned it in the source code, because it is contributed to the same projekt. If you think it is nevertheless necessary, I can add it.

### DIFF
--- a/addons/scrapers/ard
+++ b/addons/scrapers/ard
@@ -14,7 +14,7 @@ _ard_get_thumbnail_width () {
 
 _ard_correct_ytdl_pref () {
 	ytdl_pref="best"
-	print_warning "bestvideo ist not supported for ARD, fallback to best\n"
+	[ $__is_fzf_preview -eq 0 ] && print_warning "ytdl_pref $1 ist not supported for ARD, fallback to best\n"
 }
 
 scrape_ard () {
@@ -44,9 +44,13 @@ scrape_ard () {
 }
 
 on_startup_ard () {
-	# This is neccessary because ytdl only plays the audio when bestvideo is chosen (as of 2022-03-28)
-	awk \
+	# With ytdl_pref="" or ytdl="bestvideo" only audio is played (as of 2022-03-28), this fixes this:
+	if [ -z "$ytdl_pref" ]; then
+		_ard_correct_ytdl_pref "empty string" 
+	else
+		awk \
 		-v string="$ytdl_pref" \
 		-v substring="bestvideo" ' \
-		BEGIN { if (match(string, substring)) {exit 1} }' || _ard_correct_ytdl_pref
+		BEGIN { if (match(string, substring)) {exit 1} }' || _ard_correct_ytdl_pref "bestvideo"
+	fi
 }
--- a/addons/scrapers/ard
+++ b/addons/scrapers/ard
@@ -34,7 +34,7 @@ scrape_ard () {
 			thumbs: (.images.aspect16x9.src | gsub("{width}"; "'"$_thumbnail_width"'")), 
 			duration: "\(.duration / 60 | floor):\(pad_left(2; .duration % 60))",
 			date: "\(.broadcastedOn | fromdateiso8601 | strftime("%Y-%m-%d"))",
-			description: .longSynopsis,
+			description: .show.synopsis,
 		}]' < "$_tmp_json" > "$output_json_file"
 	
 }
--- a/addons/scrapers/ard
+++ b/addons/scrapers/ard
@@ -13,7 +13,7 @@ _ard_get_thumbnail_width() {
 }
 
 scrape_ard () {
-	search=$(echo "$1" | sed 's/ /+/')
+	search=$(echo "$1" | sed 's/ /+/g')
 	output_json_file="$2"
 	search_url="https://api.ardmediathek.de/search-system/mediathek/ard/search/vods?query=$search&pageNumber=0&pageSize=$((20 * pages_to_scrape))"
 	_tmp_json="${session_temp_dir}/ard_search.json"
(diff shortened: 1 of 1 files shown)


addons/scrapers/ard
```diff
@@ -0,0 +1,40 @@
...
+		high) echo "1920" ;;		# Full HD
+		default) _ard_get_thumbnail_width "high" ;;
+		medium) echo "960" ;; 		# 540p
+		sddefault) _ard_get_thumbnail_width "medium" ;;
+		*) _ard_get_thumbnail_width "default" ;;
+	esac
+}
+
+scrape_ard () {
+	search=$(echo "$1" | sed 's/ /+/')
```
Euro20179, line 16: This should probably be `search=$(echo "$1" | sed 's/ /+/g')`

addons/scrapers/ard
```diff
@@ -0,0 +1,40 @@
...
+			scraper: "ard", 
+			ID: .id,
+			url: "https://www.ardmediathek.de/video/\(.id)",
+			title: .longTitle,
+			channel: .show.title,
+			thumbs: (.images.aspect16x9.src | gsub("{width}"; "'"$_thumbnail_width"'")), 
+			duration: "\(.duration / 60 | floor):\(pad_left(2; .duration % 60))",
+			date: "\(.broadcastedOn | fromdateiso8601 | strftime("%Y-%m-%d"))",
+			description: .show.synopsis,
+		}]' < "$_tmp_json" > "$output_json_file"
```
Euro20179, line 38: This should probably use `>>` instead of `>` incase the user picks multiple scrapers.

Euro20179:
> With the default video player, it opens only the audio. With following function in conf.sh it opens the video for me
```sh
video_player () {
	mpv "$@"
}
```

Not sure what to do about this. You could add a function in the scraper like this:
```sh
on_startup_ard () {
    ytdl_pref="best"
}
```
Because I think the problem is that the ytdl_pref is defaulting to audio only, because there's no "bestvideo" available for the website, so we can just change it to best, and mpv will do it for us.

Xystel:
Thank you for the review! I corrected the two lines.

> Not sure what to do about this. You could add a function in the scraper like this:
> [...]
> If you don't want to do that, in scrape_arp, you could add something like [ "$ytdl_pref" != "best" ] && print_warning "the arp scraper works best if ytdl_pref is set to best"

I decided to go with the solution with the `on_startup_ard` function, because it takes away complexity from the end user. I first wrote a hint into that function with `print_warning` to inform the user. But that ended up in the preview window, so I removed it.
```sh
on_startup_ard () {
	# This is neccessary because ytdl only plays the audio when bestvideo is chosen (as of 2022-03-28)
	case "bestvideo" in
		*$ytdl_pref*) ytdl_pref="best" ;;
	esac
```
I also customized the function, to only apply when `$ytdl_pref` contains `bestvideo`. This respects the user choice, e. g. if the user wants starts the script with `--ytdl-pref=worst`.

Euro20179:
> But that ended up in the preview window, so I removed it.

Interesting, I should probably fix that, for now, if you still want to add  it you can put `[ $__is_fzf_preview -eq 0 ] && print_warning ...`

Also nice customization

Xystel:
> Also nice customization

Thanks, but I noticed that it had a bug. It was (for whatever reason) also triggered for `ytdl_pref=best` is given. That actually don't change the outcome, but it does not seem right to me, so I switched to an awk solution. In my tests, this solution was only triggered by `bestvideo`. 

Also I noticed that the video isn't played when `ytfl_pref` is empty, so the scraper now checks that too.

While implementing this I noticed another strange behavior. The implementation is now, that the `ytdl_pref` is filled with `best` when it is an empty string. When I'm not using the test `[ $__is_fzf_preview -eq 0 ]`, then I got the warning printed in the fzf window, even if I call ytfzf with `--ytdl-pref=best`. How can this be? It seems like `on_startup_ard` is sourced again in the fzf preview window, while `ytdl_pref` is an empty string. You can reproduce this when you edit `_ard_correct_ytdl_pref` to following content:
```sh
_ard_correct_ytdl_pref () {
	echo "$ytdl_pref"
	ytdl_pref="best"
	print_warning "ytdl_pref $1 ist not supported for ARD, fallback to best\n"
	echo "$ytdl_pref"
}
```
I'm not sure, but it seems to be an issue with ytfzf. Should I open an issue?

But as you said, for now this is fixed in this scraper with `[ $__is_fzf_preview -eq 0 ]`.

Xystel:
> I cannot fix it right now though.
No stress, take your time. 
> You can if you want, but I already know about it, and I should be able to fix it [...]
Ok, if it isn't necessary I avoid the organization overhead.

Euro20179:
> No stress, take your time.

Fixed. `on_startup_*` is no longer run in the fzf menu, i may add something like `on_new_item_selected_fzf_*` because that functionality technically worked before and now it doesn't.

In addition i changed it so that scrapers are sourced after `ytdl_pref` is set instead of before, this way it's the scrapers job to handle user preference, in case the scraper wants to modify it, like yours does.

PR #522, opened 2022-04-06
Fixes related to displaying video data

gardockt:
### Summary
1. Fixed displaying details of video with title containing quotes - when parsing JSON string object to text, current approach is to find first two quotes in the object; this PR assumes that closing quote is the last one in parsed line, using the fact that sed's `*` matches greedily
2. Fixed displaying escaped backslashes in search results - jq's `@tsv` filter escapes backslashes, which is not necessary here; alternative that does not do that is `join("\t")`
3. Fixed issues with videos that occur multiple times in search results due to paging (repeated details, thumbnail preview not updating)
   - note that this PR removes `unique_by(.ID)` filter as I do not see any reason to leave it (after filtering by URL we can assume that objects left are equal), but it can be restored if I am mistaken

### Screenshots
1.
   - before:
before_1
   - after:
after_1
2.
   - before:
before_2
   - after:
after_2
3.
   - before:
before_3
(it's worth noting that thumbnail of a previous video is shown, and "no image found" error is displayed behind it)
   - after:
after_3

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -1298,7 +1298,7 @@ download_thumbnails () {
 }
 
 get_video_json_attr () {
-	sed -n 's/^[[:space:]]*"'"$1"'":[[:space:]]*"\([^\n]*\)",*/\1/p' <<EOF | xargs -0 printf '%b'
+	sed -n 's/^[[:space:]]*"'"$1"'":[[:space:]]*"\([^\n]*\)",*/\1/p' <<EOF | sed 's/\\\([\\"]\)/\1/g'
 $_correct_json
 EOF
 }
--- a/ytfzf
+++ b/ytfzf
@@ -1298,7 +1298,7 @@ download_thumbnails () {
 }
 
 get_video_json_attr () {
-	sed -n 's/^[[:space:]]*"'"$1"'":[[:space:]]*"\([^"\n]*\)",*/\1/p' <<EOF
+	sed -n 's/^[[:space:]]*"'"$1"'":[[:space:]]*"\([^\n]*\)",*/\1/p' <<EOF | xargs -0 printf '%b'
 $_correct_json
 EOF
 }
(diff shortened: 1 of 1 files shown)


Euro20179:
> note that this PR removes unique_by(.ID) filter as I do not see any reason to leave it (after filtering by URL we can assume that objects left are equal), but it can be restored if I am mistaken

If there are duplicates it will select both when filtering by url, so we should keep it.

ytfzf
```diff
@@ -1537,7 +1537,7 @@ preview_img () {
 	url=${line##*|}
 
 	#make sure all variables are set{{{
-	_correct_json=$(jq -r --arg url "$url" '[.[]|select(.url==$url)]|unique_by(.ID)[0]' < "$video_json_file")
+	_correct_json=$(jq -nr --arg url "$url" '[inputs[]|select(.url==$url)][0]' < "$video_json_file")
```
Euro20179, line 1540: What exactly is this doing, I see from the manual that -n puts no input into the filter, and i see that `inputs` "outputs all remaining inputs, one by one". Which inputs exactly is it outputting.

In addition how does this fix the problem, it looks like it's doing the same thing?

Euro20179:
> (it's worth noting that thumbnail of a previous video is shown, and "no image found" error is displayed behind it)

Which thumbnail viewer are you using? Something is probably going wrong in the `preview_no_img` function

gardockt:
> > note that this PR removes unique_by(.ID) filter as I do not see any reason to leave it (after filtering by URL we can assume that objects left are equal), but it can be restored if I am mistaken
> 
> If there are duplicates it will select both when filtering by url, so we should keep it.
>
We are already selecting only the first object using `[0]`.

> Which thumbnail viewer are you using?
>
I'm using ueberzug 18.1.9.

ytfzf
Euro20179: ahhh, thats really cool. I get it.

ytfzf
```diff
@@ -1298,7 +1298,7 @@ download_thumbnails () {
 }
 
 get_video_json_attr () {
-	sed -n 's/^[[:space:]]*"'"$1"'":[[:space:]]*"\([^"\n]*\)",*/\1/p' <<EOF
+	sed -n 's/^[[:space:]]*"'"$1"'":[[:space:]]*"\([^\n]*\)",*/\1/p' <<EOF | xargs -0 printf '%b'
```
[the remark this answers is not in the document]
gardockt: Done - changed to `sed 's/\\\([\\"]\)/\1/g'`, which also unescapes backslashes.

Euro20179:
> I'm using ueberzug 18.1.9.

Yeah for other viewers this works properly, I'm gonna try making a different system, where instead of just displaying "no image found", it will try to load a default image if the user has it in their config. If it can't load it, then it displays no image found.

PR #538, opened 2022-05-16
Solve issue ytfzf/issues#537

qoheniac:


Euro20179:
This is cool, but I think it would make more sense to exit `ytfzf` when ctrl-c is pressed in mpv instead of just quitting normally.
ctrl-c makes the exit code 4.

qoheniac:
Makes sense. This should work.

PR #540, opened 2022-05-24
Fix typos in manuals

gpxlnx:
Hi,

I found some typos in your manuals and it would be interesting to see if your mistakes can be corrected.

Thanks for your work!!!


Euro20179:
In the future please use the development branch for changes.

However this change is very minor, so I will merge it.

PR #542, opened 2022-05-31
Add option for getting channel feed URL

txgk:
It is quite annoying to search channel id in the page source for getting feed URL. Why not use my favorite youtube script for this?

### DIFF
--- a/docs/man/ytfzf.1
+++ b/docs/man/ytfzf.1
@@ -64,6 +64,10 @@ This can also be set in the config file with
 .BR \-f ", " \-\-formats
 Show available formats before proceeding.
 .TP
+.TP
+.BR \-F " " \fIchannel-url\fR ", " \-\-feed " " \fIchannel-url\fR
+Get Atom feed URL of the channel.
+.TP
 .BR \-\-format\-selection=\fIscreen\fR
 The format selection screen type to use.
 .PP
--- a/ytfzf
+++ b/ytfzf
@@ -433,6 +433,7 @@ usage () {
         -d			Download the selected video(s)
         -m			Only play audio
         -f                      Select a video format before playing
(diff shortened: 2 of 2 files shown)


Euro20179:
There's already an option called `--channel-url` that you can use to get the "real" link to a channel (which includes the id), so this feels a bit unnecessary, especially considering how specific it is.

If you added the below to `~/.config/ytfzf/conf.sh` you can use `ytfzf --feed="a-url-to-a-channel"`

```sh
on_opt_parse_feed (){
    link=$(_get_real_channel_link "$optarg")
    echo "https://www.youtube.com/feeds/videos.xml?channel_id=${link##*/}"
    exit 0
}
```

I believe that if this function supported multiple sites, eg: reddit, it would be a lot more useful. However I think that either way, having it in the main script is a bit excessive, and works better as an extension (I'll try to write it)

Also please base your changes on the development branch.

Euro20179:
I have created a scraper that does pretty much what this does.

Here is the link to the file: https://github.com/pystardust/ytfzf/blob/development/addons/scrapers/feed-url

It's basically just an improved version of what this is.

You can do `ytfzf -c feed-url youtube/pewdiepie r/linux` and it will give you the rss feeds for pewdiepie's channel, and r/linux, in addition it sends a ping to the rss links to make sure they exist.

txgk:
Thank you, this is pretty useful. But the flaw is that `_get_real_channel_link` function is not error resistant - it returns gibberish for invalid channels and fails when youtube link doesn't have `www` prefixed.

Euro20179:
>  it returns gibberish for invalid channels

Shouldn't be too hard to fix

> And fails when youtube link doesn't have www prefixed

Again shouldn't be too hard.

However we'll just have to see, and feel free to give fixing these a shot

Issue #549, opened 2022-07-01
[BUG]: Is Vid Puffyan DOWN???


XXCryp70XX:
For those who are having an issue with scraping, because https://vid.puffyan.us seemed to be down at this moment hopefully soon they will fix the issue.

The issue: 
YouTube (with https://vid.puffyan.us) (s, pg: 1)
Youtube failed to load website
Nothing was scraped

![image](https://user-images.githubusercontent.com/94202188/176936613-7783c975-9951-4c59-a399-5c9f15934be5.png)

I got this result from a website:
![image](https://user-images.githubusercontent.com/94202188/176936911-4494b466-083c-4cfe-96bd-bdf76c673083.png)
![image](https://user-images.githubusercontent.com/94202188/176937156-02a50697-0a5d-43e3-b7b6-23198e160d95.png)

The site currently: 
![image](https://user-images.githubusercontent.com/94202188/176937223-43381964-a329-4991-9b34-0096358f313c.png)

Euro20179:
You can always use a different instance other with the --ii option in 2.4 or in your config add `invidious_instance="https://instance"`

Euro20179:
Should be fixed with #550

PR #550, opened 2022-07-02
changed the individous_instance

hoowwshii:
vid.puffyan.us is down, y.com.sb works fine for me..


hoowwshii:
well, i changed the individous_instance because of #549 thanks to @Euro20179 . i tried many different mirror, but it didn't seem to work, but this y.com.sb works for me..

Euro20179:
Actually vis.puffyan.us seems to be working for me but since a lot of people are having problems i will merge this

PR #552, opened 2022-07-12
Allow for multiline env vars when generating and reading /tmp/ytfzf-env

HoffsMH:
If you want to have a multiline string anywhere in your environment such as
```shell
export FZF_DEFAULT_OPTS='--border
--bind "ctrl-y:execute-silent(echo {} | xclip)+abort"
--preview "(pistol {}) 2> /dev/null | head -500"'
```
`ytfzf` will error with
```
/usr/bin/ytfzf: line 488: export: --: invalid option
export: usage: export [-fn] [name[=value] ...] or export -p
```

because it is trying to `export`  `--bind`

(I realize I could put it all on one line but it looks neater in my zshrc this way and it works with fzf just fine) :stuck_out_tongue: 

This change set tells `env` to delimit variable definitions by the special unix null character, by calling it with `-0`. This character is much less likely to be found in an env var as it denotes the end of a string.

stuff I searched to come to this solution:
https://www.man7.org/linux/man-pages/man1/env.1.html
https://stackoverflow.com/questions/8677546/reading-null-delimited-strings-through-a-bash-loop

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -400,7 +400,7 @@ set_vars () {
     #save the ecurrent environment so that any user set variables will be saved
     if [ "$check_exists" -eq 1 ]; then
 	tmp_env=/tmp/ytfzf-env
-	env > "$tmp_env"
+	export -p > "$tmp_env"
     fi
 
     gap_space="                                                                                                                   "
@@ -484,12 +484,28 @@ set_vars () {
 
     #read from environment to reset any variables to what the user set
     if [ "$check_exists" -eq 1 ]; then
+       _vars=$(awk -F"\n" '
+BEGIN {
+    vars=""
+}
+/^[^"]+="/{
+    vars=vars "\n" $1
+}
+!/^[^"]+="/{
+    vars=vars "\t" $1
+}
+END{
+    print vars;
+}
+' < "$tmp_env")
 	while read -r variable; do
-	    export "${variable?}"
-	done < "$tmp_env"
-	rm "$tmp_env"
+           [ -z "$variable" ] && continue
+           export "${variable#export }"
+       done <<EOF
+$data
+EOF
     fi
-    unset check_exists
+    unset check_exists _vars variable
 }
 
 set_vars "${check_vars_exists:-1}"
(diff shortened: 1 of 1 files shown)


Euro20179:
I appreciate the effort, however, `env -0`, `read -d`, and starting a string with a $ are all not posix compliant as defined [here](https://pubs.opengroup.org/onlinepubs/9699919799/).

I'll try to look for a way to do the same thing but while confining to posix, but if I can't, I wouldn't mind doing this, but having the old way be a fail safe.

Euro20179:
Aight i think i've found a solution, however it's a tad slow

```
diff --git a/ytfzf b/ytfzf
index 208cd23..ae6b778 100755
--- a/ytfzf
+++ b/ytfzf
@@ -400,7 +400,7 @@ set_vars () {
     #save the ecurrent environment so that any user set variables will be saved
     if [ "$check_exists" -eq 1 ]; then
        tmp_env=/tmp/ytfzf-env
-       env > "$tmp_env"
+       export -p > "$tmp_env"
     fi
 
     gap_space="                                                                                                                   "
@@ -484,10 +484,26 @@ set_vars () {
 
     #read from environment to reset any variables to what the user set
     if [ "$check_exists" -eq 1 ]; then
+       data=$(awk -F"\n" '
+BEGIN {
+    vars=""
+}
+/[^"]+="/{
+    vars=vars "\n" $1
+}
+!/[^"]+="/{
+    vars=vars "\t" $1
+}
+END{
+    print vars;
+}
+' < "$tmp_env")
        while read -r variable; do
-           export "${variable?}"
-       done < "$tmp_env"
-       rm "$tmp_env"
+           [ -z "$variable" ] && continue
+           export "${variable#export }"
+       done <<EOF
+$data
+EOF
     fi
     unset check_exists
 }
 ```

HoffsMH:
line continuation in the variable seems to break my installation of fzf.

I'll rebase ontop of development and try to incorporate your changes thanks!

Euro20179:
Actually it's a little weird this works at all because I made an error, line 505 should be `$_vars` not `$data`

PR #571, opened 2022-09-08
feat: use invidious api to get healthy instances

### LINKED ISSUE #570: [Feature request]: Use Invidious API to select healthy instances
### Background

- [x] I have checked ytfzf(1) and ytfzf(5) and the wiki before creating this.

### Problem

Currently, there are three ways of controlling which Invidious instance is used by `ytfzf` to search for a given command-line query:

1. Using the `--ii` flag (`ytfzf(1)`)
2. Setting a user override for `invidious_instance` in `$XDG_CONFIG_HOME/ytfzf/conf.sh` (`ytfzf(5)`)
3. Changing the default instance [in the source code](https://github.com/pystardust/ytfzf/commit/0f5672c662d3856b2f1b7bb5b00ea90e680e8ed8)

The major issue with these methods is that if an instance faces downtime, users will have to _manually_ change the instance they use (or wait for hot-fixes). In the long term, this starts to become unsustainable.

Additionally, another minor issue with hard-coding an instance is that it loads that particular instance, as all users without additional configuration, will end up querying that instance every time.

### Feature
In order to solve this issue, I propose the following.
Following [libredirect](https://github.com/libredirect/libredirect)'s example, we can use the Invidious API to set a healthy instance. The process is as follows:

1. Get the list of healthy instances from the [Invidious API](api.invidious.io). This list should be cached to improve subsequent performance.
2. Parse the JSON file, and consider the top N(=10) healthy instances. Out of these N, select one at random (using `shuf()` or similar).
3. Set the value of `invidious_instance` to the value selected in 2.

Note: the `shuf()` call is simply to ensure that the load is distributed among the top N instances.

The above process then ensures that `ytfzf` auto-selects healthy instances. Additionally, we may want to refresh the cached file at some regular interval to ensure that it hasn't become stale.

### Alternative solutions to problem?
We could just carry on with making new commits whenever an instance seems to stop working, or let users make the decision of choosing a particular instance :man_shrugging: 

If this sounds like a feature that makes sense, please let me know, and I'd be happy to make a pull request!
Thanks for developing and maintaining such a wonderful piece of software :D

BSGalvan:
Should fix #570.
Quick and dirty implementation of the process explained in the issue above.
I am not well versed with `jq`, so this might be a roundabout manner of doing what needs to be done!

~TODO: need to implement logic for staleness of the cached instance list.~

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -2150,7 +2150,7 @@ parse_opt () {
         f|formats) show_formats=${optarg:-1} ;;
         H|history) scrape="history" ;;
         x|history-clear) clear_hist "${optarg:-all}"; exit 0 ;;
-        S|select) interface="scripting" && is_specific_select="2" && scripting_video_count="$optarg" ;;
+        S|select) interface="scripting" && is_specific_select="1" && scripting_video_count="$optarg" ;;
         a|auto-select) [ -z "$optarg" ] || [ "$optarg" -eq 1 ] &&  interface="scripting" && is_auto_select=${optarg:-1} ;;
         A|select-all) [ -z "$optarg" ] || [ "$optarg" -eq 1 ] && interface="scripting" && is_auto_select=${optarg:-1} && scripting_video_count='$' ;;
         r|random-select) [ -z "$optarg" ] || [ "$optarg" -eq 1 ] && interface="scripting" && is_random_select=${optarg:-1} ;;
--- a/ytfzf
+++ b/ytfzf
@@ -572,28 +572,24 @@ set_vars () {
     thumbnail_quality="high"
     sub_link_count="2"
 
-    # URL for instance list sorted by type and health
-    instances_url="https://api.invidious.io/instances.json?sort_by=type,health" 
+    # URL for instance list sorted by type, health and api (ones without API do not work!)
+    instances_url="https://api.invidious.io/instances.json?sort_by=type,health,api" 
 
     # Cache location.
-    # TODO: decide on cache interval to avoid instances.json becoming stale
     instances_file="$cache_dir/instances.json"
 
-    # If file doesn't already exist, cURL and cache it.
+    # If file doesn't already exist (or if force-refresh is requested), cURL and cache it.
+    # TODO: implement `--refresh-inv-instances` to force-refresh this list
+    # TODO: + update docs
     [ ! -f "$instances_file" ] && curl -X GET -s "$instances_url" > "$instances_file"
  
     # The pipeline does the following:
     #   - Get top 10 instance URIs
-    #   - Trim out extra symbols produced by `jq`
-    #   - Remove empty lines
+    #   - Output these on individual lines
     #   - Shuffle the list
     #   - Choose the first entry
-    invidious_instance="$(jq -r 'map(.[1] | .uri) | .[0:10]' < "$instances_file" \
-                        | tr -d '[]," ' \
-                        | sed "/^$/d" \
-                        | shuf \
-                        | head -n1)"
-
+    invidious_instance=$(jq -r 'map(.[1] | .uri) | .[:10] | join("\n")' < "$instances_file" \
+                      | shuf | head -n 1)
 
     yt_video_link_domain="https://youtube.com"
     search_sort_by="relevance" search_upload_date="" search_video_duration="" search_result_type="video" search_result_features="" search_region="US"
--- a/ytfzf
+++ b/ytfzf
@@ -2278,7 +2278,7 @@ do_an_event_function "on_post_set_vars"
 instances_url="https://api.invidious.io/instances.json?sort_by=type,health,api"
 
 # Where do we put the list of healthy instances?
-instances_file="$cache_dir/instances.json"
+: "${instances_file:="$cache_dir/instances.json"}"
 
 # If file doesn't already exist (or if force-refresh is requested), cURL and cache it.
 # CHECK: implement check for force-request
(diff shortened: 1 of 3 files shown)


ytfzf
```diff
@@ -571,7 +571,31 @@ set_vars () {
     #this comes from invidious' api
     thumbnail_quality="high"
     sub_link_count="2"
-    invidious_instance="https://y.com.sb" yt_video_link_domain="https://youtube.com"
+
+    # URL for instance list sorted by type and health
+    instances_url="https://api.invidious.io/instances.json?sort_by=type,health"
```
Euro20179, line 576: Should also sort by `api` because ones without api dont work.

ytfzf
```diff
@@ -571,7 +571,31 @@ set_vars () {
...
     thumbnail_quality="high"
     sub_link_count="2"
-    invidious_instance="https://y.com.sb" yt_video_link_domain="https://youtube.com"
+
+    # URL for instance list sorted by type and health
+    instances_url="https://api.invidious.io/instances.json?sort_by=type,health" 
+
+    # Cache location.
+    # TODO: decide on cache interval to avoid instances.json becoming stale
+    instances_file="$cache_dir/instances.json"
```
Euro20179, line 580: I also tried my hand at implementing your idea, and I avoided having an interval because it sounded messy, what I did was create a `--refresh-inv-instances` option that just refreshes the cache.

However if you think that it can be done clearly, I'm open ears.

ytfzf
```diff
@@ -571,7 +571,31 @@ set_vars () {
...
+    #   - Get top 10 instance URIs
+    #   - Trim out extra symbols produced by `jq`
+    #   - Remove empty lines
+    #   - Shuffle the list
+    #   - Choose the first entry
+    invidious_instance="$(jq -r 'map(.[1] | .uri) | .[0:10]' < "$instances_file" \
+                        | tr -d '[]," ' \
+                        | sed "/^$/d" \
+                        | shuf \
+                        | head -n1)"
```
Euro20179, line 595: I believe this can be rewritten as:

```sh
invidious_instance=$(jq -r 'map(.[1] | .uri) | .[:10] | join("\n") < "$instances_file" \
                      | shuf | head -n 1)
```

Euro20179 (changes requested):

ytfzf
BSGalvan: Your idea sounds much more sensible. It allows us to entirely sidestep having to arrive at a properly motivated caching interval, which is much cleaner.

ytfzf
BSGalvan: Yup, you're right. Code now reflects this.

ytfzf
```diff
@@ -571,7 +571,27 @@ set_vars () {
...
+    # TODO: + update docs
+    [ ! -f "$instances_file" ] && curl -X GET -s "$instances_url" > "$instances_file"
+ 
+    # The pipeline does the following:
+    #   - Get top 10 instance URIs
+    #   - Output these on individual lines
+    #   - Shuffle the list
+    #   - Choose the first entry
+    invidious_instance=$(jq -r 'map(.[1] | .uri) | .[:10] | join("\n")' < "$instances_file" \
+                      | shuf | head -n 1)
```
Euro20179, line 592: I think this should be moved to the `Post opt scrape` section around line 2293.
If the user sets `instances_file` or `cache_dir` in their config, the file will be in the wrong place because the config is read after creating the file.

Then, instead of just setting `invidious_instance` we should only set it if it's not already set, so
```sh
: "${invidious_instance:=$(jq -r 'map(.[1] | .uri) | .[:10] | join("\n")' < "$instances_file" \
                      | shuf | head -n 1)}"
```

BSGalvan: I believe it is now in the proper section. Can you confirm if this is what you meant?

Euro20179: All good, just change:

```sh
instances_file="$cache_dir/instances.json"
```
to
```sh
: "${instances_file:="$cache_dir/instances.json"}"
```

BSGalvan: Ah, yes. Fixed now.

Also, would you prefer to isolate updates for `ytfzf(1)` required to document `--refresh-inv-instances` in a separate PR, or should we club it in along with this PR?

Euro20179: Lets just do it in this PR.

ytfzf
```diff
@@ -2271,6 +2274,27 @@ do_an_event_function "on_post_set_vars"
...
+
+# Where do we put the list of healthy instances?
+: "${instances_file:="$cache_dir/instances.json"}"
+
+# If file doesn't already exist (or if force-refresh is requested), cURL and cache it.
+# CHECK: implement check for force-request
+# CHECK: added --refresh-inv-instances to optargs
+# TODO: update docs
+[ ! -f "$instances_file" ] ||
+    [ "$refresh_inv_instances" -eq 1 ] &&
```
Euro20179, line 2288: `refresh_inv_instances` should be set to 0 in the `set_vars` function, otherwise there's an error when `--rii` ins't givent.

BSGalvan: Alright, added that in too.
I've added some options to the `curl` command, to avoid overwriting cache with an empty file, in case of internet issues (which is what inspired that fix too).

Euro20179: This seems to work well, nice addition

ytfzf
```diff
@@ -2271,6 +2276,28 @@ do_an_event_function "on_post_set_vars"
...
+    echo "Fetching list of healthy invidious instances ..." &&
```
Euro20179, line 2291: This should be
```sh
print_info "Fetching list of healthy invidious instances ...\n"
```
instead of `echo ...`

Euro20179:
Nice work!

BSGalvan:
Thank you for all the help as well!

PR #576, opened 2022-09-19
Fix addons shebang.

brauliohms:
Follows fix shebangs in some files according to the packaging in Debian, because I am the maintainer of the package in Debian.

### DIFF
--- a/addons/extensions/ani-url-handler
+++ b/addons/extensions/ani-url-handler
@@ -1,3 +1,5 @@
+#!/bin/sh
+
 open_url_handler () {
     urls="$(tr '\n' ' ' < "$1")"
 
--- a/addons/scrapers/ani-category
+++ b/addons/scrapers/ani-category
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 #stolen from pystardust/ani-cli
 _decrypt_link() {
--- a/addons/scrapers/scrape_list
+++ b/addons/scrapers/scrape_list
@@ -1,3 +1,5 @@
+#!/bin/sh
+
 YTFZF_SCRAPELIST_FILE="$HOME/.config/ytfzf/scrapelist"
 
 scrape_scrape_list () {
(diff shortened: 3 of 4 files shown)


Euro20179:
The script just sources the addons so the shebang is not very important, but if I'm going to merge this,
The `#!/bin/sh` shebangs should probably all just be `#!/usr/bin/env sh`. And `scrapers/ani-category` should use `sh` not `bash`

brauliohms:
The script `scrapers/ani-category` in line 107, lintian indicates is not posix, is a bash script.

brauliohms:
Follow the lintian msg:

bash-term-in-posix-shell '(ep_start+=' [usr/share/ytfzf/addons/scrapers/ani-category:107]

This script is marked as running under /bin/sh, but it seems to use a feature found in bash but not in the SUSv3 or POSIX shell specification.

Euro20179:
According to [this](https://pubs.opengroup.org/onlinepubs/9699919799/) which I've used as a posix reference for this script, `+=` inside arithmetic expansion is posix compliant.

> Next, the shell shall treat this as an arithmetic expression and substitute the value of the expression. The arithmetic expression shall be processed according to the rules given in [Arithmetic Precision and Operations](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap01.html#tag_17_01_02_01), with the following exceptions:
> * Only signed long integer arithmetic is required.
> * Only the decimal-constant, octal-constant, and hexadecimal-constant constants specified in the ISO C standard, Section 6.4.4.1 are required to be recognized as constants.
> * The sizeof() operator and the prefix and postfix "++" and "--" operators are not required.
> * Selection, iteration, and jump statements are not supported.

From my testing += works in `dash`, `bash`, and `busybox sh`, and `oksh (openbsd ksh)`.

Either way I can change it later

brauliohms:
Thanks, and it is really very controversial this limitation of what is POSIX or not.
And keep up the good work.

PR #577, opened 2022-09-19
Improve makefile

brauliohms:
Follows improve Makefile for better compatibility with other systems. Patch according to Debian packaging, because I am the maintainer of the package in Debian.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ doc:
 	cp docs/conf.sh ${DESTDIR}${DOCDIR}
 	cp LICENSE ${DESTDIR}${LICENSEDIR}
 
-install: doc addons
+install:
 	chmod 755 ${PROG}
 	mkdir -p ${DESTDIR}${BINDIR}
 	cp ${PROG} ${DESTDIR}${BINDIR}/${PROG}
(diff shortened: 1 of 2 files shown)


brauliohms:
The current makefile not containing the 'ALL' will have a problem when doing the make command, because make is equal to the make all command and the fact of not having the all: blank, will perform the installation of the first target that is the doc: then the recommended would be to put the all: without dependencies and commands for the make install perform the installation target install.

Euro20179:
If doing just `make` runs the first target, then I think I can just move install to the top.
When I add the `all:` target, and run `make` it says `Nothing to be done for 'all'` even though ytfzf is not installed.

brauliohms:
If you put just the all: it is enough for the `make` command to do nothing in the case of YTFZF, because the `make` command is for building and in the case of ytfzf we don't need to build, we just need to install. So a good suggestion to improve the Makefile, would be to add the blank `all:`.

brauliohms:
if you put `all: install` the `make` command will run the 'install target', and this is not correct, because the `all:` is used to build a program (if it was in C it would be time to use GCC to build the binary) but in the case of YTFZF the binary is a script and does not need to be built, just installed, so you need to leave a blank build instruction with `all:` for nothing to be built. When you need to install the binary (ytfzf script) the correct is to use `make install` or `make install docs addons` (according to your need).

Euro20179:
Ok, cool, I'll reopen this pull request, but can you get rid of the `install: doc addons` part?

brauliohms:
Ok, I understood that the user has the freedom of what to install, I will remove this part of the patch.

PR #586, opened 2022-10-10
Make the Makefile 'all' target phony and default

mathisto:
`make` it so

I noticed in the README, we are encouraging users to include the docs by default.

I also noticed the makefile had an empty `all` target. GNU `make`, when invoked without an explicit target, will snag the first one it finds.

This PR simply adds the default `install` and `doc` targets the default. I updated the README to reflect this as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,14 @@ BINDIR=${PREFIX}/bin
 DOCDIR=${PREFIX}/share/doc/ytfzf
 MANDIR=${PREFIX}/share/man
 LICENSEDIR=${PREFIX}/share/licenses/ytfzf
+
 YTFZF_SYSTEM_ADDON_DIR=${PREFIX}/share/ytfzf/addons
 
-all: install doc
+.DEFAULT_GOAL := default
+
+all:
+
+default: install doc
 
 doc:
 	mkdir -p ${DESTDIR}${MANDIR}/man1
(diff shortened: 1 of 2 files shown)


Euro20179:
[Someone else said it was a bad idea](https://github.com/pystardust/ytfzf/pull/577#issuecomment-1251741426) to do this exact  thing, I think they were just saying  it's non-standard to do this, but I'm not sure.

mathisto:
[Ahh, I see.](https://github.com/pystardust/ytfzf/pull/577#issuecomment-1251741426)

Gee willikers! Boy do I not like `make`.

So, I dug in a bit on this, and I think this is referencing 'standard targets for users'. It should be noted that this is *not* a POSIX requirement. The `all` target has no special meaning to `make` whatsoever. That being said, GNU manual [does recommend](https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html#Standard-Targets) that programs use it, and additionally admonishes it NOT include the docs.

So as to keep the `all` directive intact, useless as it is, I added the `.DEFAULT_GOAL` directive which [does](https://www.gnu.org/software/make/manual/html_node/How-Make-Works.html) have special significance to `make`. This directive was added in `v3.80`. Version 3.80 just had its 20th birthday on the 3rd, so we should be good for portability, methinks.

Let me know what you think.

Euro20179:
> Version 3.80 just had its 20th birthday on the 3rd, so we should be good for portability, methinks.

Honestly you would think it would be portable, but I tried with the arch package `bmake` (bsd make) and it does nothing.

However, it also doesn't break `bmake` so I think it's fine, you would just have to do it the old way (`bmake install doc`).

Thanks for doing the research on make btw.

mathisto:
OOOF. You know, I've never even used BSD and it makes feel downright silly sometimes.
Ok, I'll pull the readme changes. Thanks for working with me on this. ✋

PR #587, opened 2022-10-11
Add required whitespace to vim modeline

mathisto:
MAGIC MODELINES

I noticed the Vim modeline was not working. I checked all my settings locally and sure enough, it wasn't being picked up because of a single space after the comment `#`. 

According to [the docs for Vim modelines](https://vimdoc.sourceforge.net/htmldoc/options.html#modeline), the space following `vim:` is optional, but the one preceding it is not. This should ensure proper detection.

This PR adds not only the required space, but goes ONE STEP FURTHER, and adds an *optional* one following the colon. This is avandt-garde code styling, I KNOW. FWIW the help doc provides an example of each, with and without the trailing space after the colon.

Additionally, modelines are a known security vulnerability. Consider removing it to protect users running a version less than `8.1.1365`.  A lot of distributions have it disabled by default. Debian adds explicit warnings. More [here](https://github.com/numirias/security/blob/master/doc/2019-06-04_ace-vim-neovim.md) if you are interested.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -19,7 +19,8 @@ assignees: ''
  - OS:
  - Terminal:
  - Ytfzf version:
-- Output of `ls -l "$(which sh)"` (if your using fish: `ls -l (which sh)`):
+- Output of `ls -l "$(which sh)"` (if you're using fish: `ls -l (which sh)`):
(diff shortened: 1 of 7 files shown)


Euro20179:
Can you reopen this and put it on the development instead of master branch.

mathisto:
OH JESUS THAT IS EMBARRASING

PR #589, opened 2022-10-11
 Use consistent whitespace across all comments

mathisto:
Just dusting

I went through with a fine tooth regex comb and manually fixed all the comments that that did not have a trailing space. It was mix and match throughout the code which is pretty normal to see on collaborative bases like this. 

As it stands, there are 8 remaining instances of a `#` without a trailing whitespace. All of these are legit code and have remain unchanged.


Euro20179:
Nice consistency fix.

PR #604, opened 2022-11-16
Support for channel links with `@`

two-six:
This allows to use command `ytfzf --channel-link ` on links like `https://www.youtube.com/@{channel_name}`

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -800,7 +800,7 @@ _get_real_channel_link () {
         https://*|*\.*)
             domain=${1#https://}
             domain=${domain%%/*}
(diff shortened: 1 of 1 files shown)


Euro20179:
I don't think this is necessary since it doesn't seem to do anything, I think youtube is preferring \@user over user ids, so I think I'll just let the channel scraper also accept \@user

Euro20179:
I made [this](https://github.com/pystardust/ytfzf/commit/493f59832caeeb6068f6611f4cd106a76a54c3eb) patch, because I tried yours and it wasn't working with something like `https://www.youtube.com/@pewdiepie`, The main difference is that i made it change `@user` to `/user/<user>` instead of `/c/<user>`, but I also made it work with just `@user` no url required.

two-six:
I see how it works now, thanks for that and this patch 😃

PR #625, opened 2023-01-14
add viewed date support

jpgichw:
this is an old altered version I had lying around on my computer, seems like there is a lot that has changed since then. still working after a bit of modification. in the text interface the viewed date only shows the date and time, not the timezone but the timezone is shown in the thumbnail interface.

extra commands: date, tac. both are part of the GNU coreutils


jpgichw:
(wrong branch)

PR #627, opened 2023-01-14
Better kitty compatibility when displaying images

greenfoo:
First of all, I understand this is a dirty hack, thus please feel free to discard it if you don't like it.

Anyway... turns out that when using `--scale-up` in kitty to make the thumbnail fill as much space as possible then, when certain conditions are met, the UI becomes "scrambled", like this:

scrambled

These conditions are not 100% clear to me, but I think it has to do with, at least, the following things:
* Running kitty inside **a tabbed frame in sway** (when I don't have tabs this does not seem to happen)
* Previewing **specific thumbnails** (maybe related to their aspect ratio?)

I can't really say what the root case of the issue is (that would be something to discuss in kitty's bug tracker) but after applying this patch I have not been able to reproduce the issue again.

Some extra notes:
* Why "-2"? I first tried with "-1" but the bug was still happening. Is "-2" enough? No idea, maybe we could use something bigger, like "-10" to play safe.
* I see something similar was done for the case when `chafa` is used. In that case "-10" was used. Should we then do the same here?

Thanks!

### DIFF
--- a/ytfzf
+++ b/ytfzf
@@ -2629,7 +2629,7 @@ EOF
 	    kitty)
 		get_ueberzug_positioning "$FZF_PREVIEW_COLUMNS" "$FZF_PREVIEW_LINES" "$fzf_preview_side"
 		kitty +kitten icat --clear --transfer-mode file > "$thumbnail_debug_log" 2>&1
-		kitty +kitten icat --place "${width}x${height}@${x}x${y}" --scale-up --transfer-mode file "$thumb_path" > "$thumbnail_debug_log" 2>&1 ;;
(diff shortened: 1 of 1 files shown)


Euro20179:
If the problem is indeed `--scale-up`, I think we should just remove the use of that option for 2 reasons:

1. it works the same without it
2. changing the width/height makes it off center.

Euro20179:
Oh interesting, I tested it and it worked fine, if it requires `--scale-up` then fair enough, we can just change the size.

Before I merge, can you base your  changes off the `development` branch instead of master, then merge to development.

PR #628, opened 2023-01-14
Better kitty compatibility when displaying images

greenfoo:
These are the same changes previously discussed in #627 but rebased on top of the `development` branch instead of `master`


Euro20179:
Looks good.

Issue #637, opened 2023-02-01
[Feature request]: re use mpv instance 


citosid:
### Background

[x] I have checked ytfzf(1) and ytfzf(5) or the wiki before creating this.

### Problem

Every time ytfzf is run, a new instance of the player is opened

### Feature

Running ytfzf from the command line will open a new window of mpv every time. However, it is possible to configure mpv to use sockets and send a command to load the new url instead of opening a new instance

```bash
echo "{ \"command\": [\"loadfile\", \"$1\"] }" | socat - "/tmp/mpv-socket"
```

### Alternative solutions to problem?

Kill the current one and open it again? I use i3 (specifically regolith) and have the mpv player configured to be sticky and small so it always shows in my desktop:

![image](https://user-images.githubusercontent.com/201430/216159063-b691ad22-303d-4d0b-af2a-8564e2f1b98f.png)

So it would always span a new window and will lost that configuration... as you can see in the image, I tried with a local file and it works. Will try to do a MR for this feature, but not sure if there is any downside to do it like that

Euro20179:
If you want to work on a pull request, i'd recommend modifying the `mpvq` addon to work across all instances of ytfzf.

Or you could just make a new addon that basically does the same thing as `mpvq` but works on all instances.

I probably wouldn't merge this feature into the main script.

citosid:
Thanks for the quick reply @Euro20179 ! You beat me to it lol. 

Could you at least review the code I have to see if you see any problem? I will look into either modifying mpvq or adding a new one.

tabletseeker:
Updated and working ytfzf project here:
https://github.com/tabletseeker/ytfzf_prime

PR #638, opened 2023-02-01
feat(mpv): use single instance of mpv

citosid:
This commit configures mpv to create an ipc-server and re-use it to playback the files so no new instances are created

Issue #637


Euro20179:
I think you only modified the `audio_player`, this won't matter if written as an addon, because your addon will be run as the url handler instead of `audio_player` or `video_player`

`pidof` is not posix, however if written as an addon, it doesn't matter, just check if it's installed.

There is one problem that may occur where if there is more than 1 mpv instance open it could pick the wrong one.

for this line of code
> echo "{\"command\":[\"loadfile\",\"${@}\", \"replace\"]}"

I would use `$*` instead of `${@}` because `${@}` might act weirdly with word splitting and stuff.

PR #673, opened 2023-04-29
Fix not being able to disable history

futureisfoss:
setting `enable_hist` and/or `enable_search_hist` to `0` should prevent saving history, at least the documentation says so but it wasn't working. This PR fixes the bug.


Euro20179:
I may end up changing a few things later, but this is good for now.

